PR #26, opened 2018-01-16
fixed index into list pmap potential field

violetguos:
the indices/ranges of ```pmap``` are not casted to integers, minor issue. Please review my pull request and changes in my cloned repository. Thank you.


AtsushiSakai:
@violetguos Hi. thank you for your contribution again 😃
I think these variable are always integers, so I think we don't need to cast it.
Because, wx, wy came from round function, and ix, iy came from range function.
What do you think it?

AtsushiSakai:
Could you please let me know your versions of python and mac OS.
In my env, it doesn't happen..

violetguos:
Sorry I was running it in Python 2.7. It is now working when compiled in Python 3.6. Thank you for your reply!

AtsushiSakai:
@violetguos OK. Thank you so much for trying my project. If there are any issues, let me know😃.

PR #35, opened 2018-01-31
Jacobian computation in  model_predictive_trajectory_generator.py

yasnem:
Mistake in computation of the Jacobian.


AtsushiSakai:
@yasnem Thank you for your contribution. You are right. I appreciate your work.

PR #44, opened 2018-03-14
Fixed occasional math domain error

daniel-s-ingram:
Within the first ten or so iterations, taking the square root of an extremely small negative number was attempted, resulting in a math domain error.


AtsushiSakai:
@daniel-s-ingram Thank you for good PR.

daniel-s-ingram:
No problem!

PR #48, opened 2018-03-22
Fixed issue with float index for Python 2

daniel-s-ingram:
The floor function in Python 2 returns a float, leading to an index error in some cases. Converted indexes explicitly to ints.

### DIFF
--- a/Mapping/raycasting_grid_map/raycasting_grid_map.py
+++ b/Mapping/raycasting_grid_map/raycasting_grid_map.py
@@ -50,7 +50,7 @@ def atan_zero_to_twopi(y, x):
 
 def precasting(minx, miny, xw, yw, xyreso, yawreso):
 
-    precast = [[] for i in range(round((math.pi * 2.0) / yawreso) + 1)]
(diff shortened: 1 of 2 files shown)


AtsushiSakai:
@daniel-s-ingram Thank you so much for contributing again.
Let me ask one thing. 
Do you think the repository should work in Python2?
I don't want to do that... because the code becomes complicated.
The main purpose of this repo is for understating the algorithm's basic idea.

daniel-s-ingram:
I think compatibility with Python 2 would be great to have for people who want to learn the algorithms but don't have Python 3 or its libraries installed due to potential conflicts with their own projects. If it's okay with you, I can keep fixing small issues like this without making the code any more complicated beyond the small changes I've been making.

AtsushiSakai:
@daniel-s-ingram OK. I think this PR doesn't make the code so complicated. I will merge this PR. Thank you for your work again!!

daniel-s-ingram:
Thank you - great repo, by the way!

PR #53, opened 2018-03-27
analytic jerk calculation in frenet_optimal_planner.py

jwdinius:
The previous implementation of the frenet optimal planner used a finite difference of acceleration for the jerk calculation.  I've updated the planner to use analytical jerk calculations using the time derivatives of polynomials.


AtsushiSakai:
@jwdinius Thank you for great PR again. I appreciate your work 😃.  I added your name to README.md as an author of this repo. Thank you for your contribution again!!.

PR #62, opened 2018-06-03
Update extended_kalman_filter.py

minxuanjun:
Jacoby's formula is wrong,for lacking u


AtsushiSakai:
Thank you for your contribute. Yes, you are right. 
The jacobian matrix seems to be wrong. But I found an another problem.
Please check latest version of the code.

PR #63, opened 2018-06-03
Update extended_kalman_filter.py

minxuanjun:
Jacoby's formula is wrong

### DIFF
--- a/Localization/extended_kalman_filter/extended_kalman_filter.py
+++ b/Localization/extended_kalman_filter/extended_kalman_filter.py
@@ -87,7 +87,7 @@ def jacobF(x, u):
         [1.0, 0.0, -DT * u1 * math.sin(yaw), DT * u1 * math.cos(yaw)],
         [0.0, 1.0, DT * math.cos(yaw), DT * math.sin(yaw)],
         [0.0, 0.0, 1.0, 0.0],
-        [0.0, 0.0, 0.0, 1.0]])
+        [0.0, 0.0, 0.0, 0.0]])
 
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@minxuanjun Thank you for your contribution.
Could you please explain why the element should be 0?
I think v_{t+1} = v{t}, so dv(t)/dv = 1.0.

minxuanjun:
@AtsushiSakai , because your matrix F[3,3]==0

minxuanjun:
@AtsushiSakai  I think x and u are different state variable , d(f(x,u))/dx  and d(f(x,u))/dx should be separated.

AtsushiSakai:
@minxuanjun  Based on wikipedia article, the state equation f(x,u) seems to include both matrixs. Extended Kalman filter - Wikipedia https://en.wikipedia.org/wiki/Extended_Kalman_filter 
And the jacobian is differentiated by both current x and u.
Do you have any reference about your idea?

minxuanjun:
@AtsushiSakai  I reference the chapter7 of Probabilistic robot。

AtsushiSakai:
@minxuanjun I think the difference came from noise modeling. In my case, I assume that gaussian noise is added to state, not input for simplicity. In the Probabilistic Robotics, they assume noise is added to input.

minxuanjun:
@AtsushiSakai thank you.

AtsushiSakai:
@minxuanjun Thank you so much too for taking your time. I appreciate good disquisition with you.

PR #66, opened 2018-06-07
Adding Informed RRT Star Path Planner

karanchawla:
Based on the current implementation of the RRT Star planner in this repo, this PR adds Informed RRT Star based on the following research paper: https://arxiv.org/pdf/1404.2334.pdf

This method basically restricts the sampling space to an ellipse and gives a more optimal solution as compared to RRTStar.


AtsushiSakai:
@karanchawla Interesting!!. I will check your new code. Thank you.

karanchawla:
Do you think it'll be better if all the sampling-based planners are refactored into a separate sub-library of their own? I think it'll be better for the users to make sense of the code and also easier to refactor and integrate with other projects.

PR #69, opened 2018-06-14
Batch Informed Trees

karanchawla:
Reference: https://arxiv.org/pdf/1405.5848.pdf

Implements the ^paper. Uses a combination of graph search techniques to delay connection of nodes(since that is an expensive operation, mostly because of collision checking) and uses informed heuristics to intelligently sample the free space. 

Possible improvements: Performance and pruning the tree(Authors mention it's not important and is an expensive operation) Hence, this part is not included in the current PR.


AtsushiSakai:
@karanchawla Thanks again!! I will check it later.

PR #75, opened 2018-07-03
Astar logic update

tsaoyu:
Correct the logic on successor node cost. The *f* cost the successor node should be *g* cost of successor node plus *h* cost to the **goal**. Here your *g* cost is the same as tentative (*t*) cost which is correct, however the *f* cost is not. More see in [pseudocode](http://mat.uab.cat/~alseda/MasterOpt/AStar-Algorithm.pdf)<sup>1</sup> and [psedocode](https://gist.github.com/damienstanton/7de65065bf584a43f96a#file-astar-txt-L25)<sup>2</sup>. 

Other correct are small typo corrections.

<sup>1</sup>: line 16

<sup>2</sup>: line 25

### DIFF
--- a/PathPlanning/AStar/a_star.py
+++ b/PathPlanning/AStar/a_star.py
@@ -27,7 +27,7 @@ class Node:
         return str(self.x) + "," + str(self.y) + "," + str(self.cost) + "," + str(self.pind)
 
 
-def calc_fianl_path(ngoal, closedset, reso):
+def calc_final_path(ngoal, closedset, reso):
     # generate final course
     rx, ry = [ngoal.x * reso], [ngoal.y * reso]
     pind = ngoal.pind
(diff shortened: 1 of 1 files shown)


PathPlanning/AStar/a_star.py
```diff
@@ -105,10 +105,10 @@ def a_star_planning(sx, sy, gx, gy, ox, oy, reso, rr):
             if tcost >= node.cost:
                 continue  # this is not a better path
 
-            node.cost = tcost
-            openset[n_id] = node  # This path is the best unitl now. record it!
+            node.cost = tcost + calc_heuristic(node, ngoal)
```
AtsushiSakai: @tsaoyu Is it correct?

tsaoyu: This is correct only if `node.cost` refers to *f* cost. I think I am wrong here, cost here is *g* cost.

tsaoyu:
Although it was a shame to admit there is some flaw in my logic, it is still interesting to explore what happens here and why it didn't fail to find the shortest path. The reason is the when heuristic distance was added twice on *g* cost, it essentially means the weight in heuristic function was doubled. 
In A* search, the heuristic function doesn't have to be the exact estimation, also leads to the finding of the shortest path.

PR #78, opened 2018-07-08
Add \eta^3 spline path to PathPlanning

jwdinius:
I found an interesting path generator that is geometrically continuous up to order 1 (the derivative is guaranteed continuous).  The reference is [here](https://ieeexplore.ieee.org/document/4339545).  I was able to code up the algorithm and recreate the test case presented.  I thought that you might like to add it to your repo.


AtsushiSakai:
@jwdinius Thank you for great PR!!

PR #80, opened 2018-07-10
Update the pi_2_pi function to a use of a modulo instead of two while.

AlexisTM:
Update the `pi_2_pi` function from two while implementation to a modulo.

Initial code: 

```python
    while(angle > math.pi):
        angle = angle - 2.0 * math.pi

    while(angle < -math.pi):
        angle = angle + 2.0 * math.pi

    return angle
```

New implementation:

```python
    return (angle + math.pi) % (2*math.pi) - math.pi
```


AtsushiSakai:
@AlexisTM Thank you for good PR!!

PR #82, opened 2018-07-11
Fix spelling mistakes in README.md

herallius:
- a object -> an object
- a Extended Kalman Filter -> an Extended Kalman Filter
- cross -> crosses
- same -> the same
- It's -> Its
- means -> mean
- domein -> domain
- being patron -> being a patron


AtsushiSakai:
@herallius Thank you!!

Issue #83, opened 2018-07-12
Performance increase 


AlexisTM:
https://github.com/AtsushiSakai/PythonRobotics/blob/8652a4b7d1cc12682829b8929b528ef9b7cefb0f/PathTracking/lqr_steer_control/lqr_steer_control.py#L139-L156

In the following function, the line 143 is more expensive than it could be:
* `abs` is not required: result of SQRT and result of sum of squared values is always positive
* `sqrt` is expensive, while the logic is the same without using it

Therefore, I would do the following optimization (replacing 143 to 147):

```python
     d = [idx ** 2 + idy ** 2 for (idx, idy) in zip(dx, dy)] 

     mind = min(d) 

     ind = d.index(mind) 

     mind = math.sqrt(mind)
```

AtsushiSakai:
@AlexisTM  Good idea. Can you create a PR?

PR #84, opened 2018-07-12
Optimize calc_nearest_index

AlexisTM:
Followup of #83.

This PR is calling `math.sqrt` once in the `calc_nearest_index` function instead of `len(points)` times. 

Note: In `model_predictive_speed_and_steer_control.py` replaced `min_d`  by `mind`


AtsushiSakai:
@AlexisTM Thank you for your PR. I added you as an author in README.

AlexisTM:
You're welcome ;)

PR #92, opened 2018-07-29
add environment.yml to setup conda environment

jwdinius:
This PR addresses the feature request I just created.  Users who prefer to use conda can setup their environments by navigating to the root of the PythonRobotics repo and typing: 

```
conda env create -f environment.yml
```

Then, to use the newly created `python_robotics` environment, they can type (from any directory):

```
source activate python_robotics
```

Also included in this PR (by mistake, but I was going to commit it anyways), is logic to compute the segment length in the `eta3_spline_path.py` code that was recently added.


AtsushiSakai:
@jwdinius Great!! Thank you.

PR #96, opened 2018-08-23
Fix small doc string typo

shubh-agrawal:
Also, Line 29 is difficult to understand. Could it be written as "Beta is the angle between the vector connecting robot's position to goals position and goal's orientation"? Please let me know if it gives the wrong idea.


AtsushiSakai:
@shubh-agrawal Thank you!!

PR #109, opened 2018-10-04
Jupyter notebook for planar two-link inverse kinematics

daniel-s-ingram:
I plan to add some comments to the code and an explanation of the other possible solution, but how does this look so far?


AtsushiSakai:
@daniel-s-ingram Oh, great. It looks perfect. It must be helpful for all users!!.

PR #110, opened 2018-10-20
Update doc for arm navigation

araffin:
- added arm navigation with obstacle avoidance to html doc


AtsushiSakai:
@araffin Thanks!!

PR #112, opened 2018-10-21
modify stanley controller

wlbksy:
I'm trying to separate some utils into common area.

At some point, codes related to car model should be extracted too, such as unicycle, bicycle and so on.

### DIFF
--- a/PathTracking/stanley_controller/stanley_controller.py
+++ b/PathTracking/stanley_controller/stanley_controller.py
@@ -1,31 +1,33 @@
 """
-
 Path tracking simulation with Stanley steering control and PID speed control.
 
+DOI: 10.1002/rob.20147
+
 author: Atsushi Sakai (@Atsushi_twi)
-
 """
-from __future__ import division, print_function
-
 import sys
-
-sys.path.append("../../PathPlanning/CubicSpline/")
 
 import matplotlib.pyplot as plt
 import numpy as np
 
+sys.path.append("../../PathPlanning/CubicSpline/")
+
 import cubic_spline_planner
+from utils_math import normalize_angle
+from utils_unit import kph2mps, mps2kph
+
 
 k = 0.5  # control gain
 Kp = 1.0  # speed propotional gain
 dt = 0.1  # [s] time difference
 L = 2.9  # [m] Wheel base of vehicle
-max_steer = np.radians(30.0)  # [rad] max steering angle
+max_steer = np.deg2rad(30.0)  # [rad] max steering angle
 
 show_animation = True
 
 
-class State(object):
+
+class State:
     """
     Class representing the state of a vehicle.
 
@@ -57,8 +59,9 @@ class State(object):
         self.x += self.v * np.cos(self.yaw) * dt
         self.y += self.v * np.sin(self.yaw) * dt
         self.yaw += self.v / L * np.tan(delta) * dt
-        self.yaw = normalize_angle(self.yaw)
         self.v += acceleration * dt
+
+        self.yaw = normalize_angle(self.yaw)
 
 
 def pid_control(target, current):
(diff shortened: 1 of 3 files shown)


wlbksy:
Besides, I see this repository uses `np.matrix` instead of `np.array`.
I'd like to know that would you permit if I change it to `np.array`

AtsushiSakai:
@wlbksy Thank you for your PR again. I appreciate your help.

AtsushiSakai:
@wlbksy using np.array instead of np.matrix is good idea!! I agree that.

AtsushiSakai:
@wlbksy however, I don't think it is good to make util files into common area. I want to keep each script independently.

wlbksy:
I broke the unit test. 

Since you mentioned the idea here is to make each script independently, I'll remove the utils in this PR.

There are many duplicate codes on normalize angle to [-pi, pi] and so on,  what's your view of them?

AtsushiSakai:
@wlbksy So, could you please keep all codes inside the stanley_controller.py? and also it's test was broken because of import problem. could you please fix it?
see: [Job \#504\.1 \- AtsushiSakai/PythonRobotics \- Travis CI](https://travis-ci.org/AtsushiSakai/PythonRobotics/jobs/444241333#L2616)

AtsushiSakai:
@wlbksy 
>There are many duplicate codes on normalize angle to [-pi, pi] and so on, what's your view of them?
I know it. I think it is ok to leave it now, because many users just copy the script and using it.

PathTracking/stanley_controller/stanley_controller.py
```diff
@@ -147,15 +162,14 @@ def main():
...
+    cx, cy, cyaw, _, _ = cubic_spline_planner.calc_spline_course(ax, ay, ds=0.1)
 
-    target_speed = 30.0 / 3.6  # [m/s]
+    target_speed = kph2mps(30.0)  # [m/s]
 
     max_simulation_time = 100.0
 
     # Initial state
-    state = State(x=-0.0, y=5.0, yaw=np.radians(20.0), v=0.0)
+    state = State(x=-0.0, y=5.0, yaw=np.deg2rad(20.0), v=0.0)
```
AtsushiSakai: @wlbksy One question. why are you using deg2rad instead of radians?

PathTracking/stanley_controller/stanley_controller.py
```diff
@@ -128,17 +139,21 @@ def calc_target_index(state, cx, cy):
...
 
     # Search nearest point index
-    dx = [fx - icx for icx in cx]
-    dy = [fy - icy for icy in cy]
-    d = [np.sqrt(idx ** 2 + idy ** 2) for (idx, idy) in zip(dx, dy)]
-    error_front_axle = min(d)
-    target_idx = d.index(error_front_axle)
-
-    target_yaw = normalize_angle(np.arctan2(fy - cy[target_idx], fx - cx[target_idx]) - state.yaw)
+    squared_error_of_front_axle = np.inf
```
AtsushiSakai: @wlbksy I suppose you should not use for loop in Python. I think the list comprehension and min function is faster than it. what do you think?

PathTracking/stanley_controller/stanley_controller.py
wlbksy: They're equivalent. However, `deg2rad` is introduced since 1.13 version of `numpy`, its name is more clear than `radians`.

AtsushiSakai: oh. I see. You are right. We should use deg2rad.

PathTracking/stanley_controller/stanley_controller.py
wlbksy: I just tested and You're right. I'll revert this. Sorry for the trouble

AtsushiSakai: You don't need to say sorry : ).

AtsushiSakai: When you commit it, I will merger your PR.

wlbksy:
Thought this was ready to merge.

PR #120, opened 2018-11-02
small bugs on linear discrete model of MPC

caseypen:
In linear discrete model of C matrix, there should be a negative sign on the last item in the vector. In the code, it also misses. However, this missing does not affect the performance.


AtsushiSakai:
#199

AtsushiSakai:
@caseypen Thank you for you PR. Yes, you are totally right!!.

caseypen:
Thanks for your nice repository! It's my pleasure to have little
contribution.

Best

Chen

PR #123, opened 2018-11-03
Update a_star.py to correct 2 TypeErrors

aleonnet:
TypeError: range() integer end argument expected, got float
TypeError: list indices must be integers, not float


AtsushiSakai:
@aleonnet Hi. Thank you for your PR. Which the version of Python are you using?

aleonnet:
Hi! Version 2.7.15

PR #126, opened 2018-11-04
Fixing motion calculation and heading cost calculation

debbynirwan:
1. Fixed motion calculation
    Current motion calculation uses current angle to calculate the new x,y. It is not correct because 2 different angular speeds will result with the same x and y. The angular speed should affect the end position.

2. Rather than using distance to the goal, as proposed by the paper the heading is used to calculate the error

### DIFF
--- a/PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
+++ b/PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
@@ -35,9 +35,9 @@ class Config():
 def motion(x, u, dt):
     # motion model
 
+    x[2] += u[1] * dt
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@debbynirwan Thank you for your PR.
> because 2 different angular speeds will result with the same x and y. 
I'm sorry I cannot understand this. What do you mean it?

debbynirwan:
@AtsushiSakai, let me give an example. If we have 2 different sets of speeds, let's say:
v = 10 m/s, w = 10 rad/s
v = 10 m/s, w = 20 rad/s

You should have two different x(x[0]), y(x[1]). But in your current code, motion() will give you same x and y. 
To fix this, we can simply calculate x[2] first.

PR #127, opened 2018-11-07
fix EKF

y011d4:
* In `efk_estimation`, the prediction step of `PPred` seems to be an typo (matrix multiplication of `np.array` should be described by not `*` but `np.dot`).
* If I understand correctly, the calculation of `xPred` should not contain noise. So not `ud` but `u` should be used.
* In `observation`, noise in control is described as `Rsim` and one in observation is as `Qsim`, while in `efk_estimation`  `Q` and `R`. I think those notations should be unified (I adopted the same notation as the book, *Probabilistic Robotics*).

The comments below, about which I fixed nothing at all, contains my questions about the simulation implemented in EKF algorithm. I have no confidence because I'm just a beginner for this field.

* Aren't roles `xTrue` and `xDR` opposite? The true trajectory is determined by the control including noise.
* Accordingly the position observation should be changed so that it detects the `xDR` plus noise.

### DIFF
--- a/Localization/extended_kalman_filter/extended_kalman_filter.py
+++ b/Localization/extended_kalman_filter/extended_kalman_filter.py
@@ -53,14 +53,14 @@ def observation(xTrue, xd, u):
 def motion_model(x, u):
 
     F = np.array([[1.0, 0, 0, 0],
-                   [0, 1.0, 0, 0],
-                   [0, 0, 1.0, 0],
-                   [0, 0, 0, 0]])
+                  [0, 1.0, 0, 0],
+                  [0, 0, 1.0, 0],
+                  [0, 0, 0, 0]])
 
     B = np.array([[DT * math.cos(x[2, 0]), 0],
-                   [DT * math.sin(x[2, 0]), 0],
-                   [0.0, DT],
-                   [1.0, 0.0]])
+                  [DT * math.sin(x[2, 0]), 0],
+                  [0.0, DT],
+                  [1.0, 0.0]])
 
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@y011d4 Hi. Thank you for your PR.

AtsushiSakai:
I agree
> In efk_estimation, the prediction step of PPred seems to be an typo (matrix multiplication of np.array should be described by not * but np.dot).

and 
> In observation, noise in control is described as Rsim and one in observation is as Qsim, while in efk_estimation Q and R. I think those notations should be unified (I adopted the same notation as the book, Probabilistic Robotics).

However, I don't agree this:
>If I understand correctly, the calculation of xPred should not contain noise. So not ud but u should be used.
Because, the EKF estimation can only use ud (input with noise) because all sensor has noise.

y011d4:
> However, I don't agree this:

  >> If I understand correctly, the calculation of xPred should not contain noise. So not ud but u should be used.

> Because, the EKF estimation can only use ud (input with noise) because all sensor has noise.
> What do you think?

I thought that `u` is the control value we want to apply and `ud` is the exact control value the GOD only knows, so we cannot obtain `ud`.
But it seems that the value `ud` can be obtained somehow. I took a mistake about considered situations.
Thank you for your kindful comments!

I fixed some as you pointed out.

AtsushiSakai:
@y011d4 > But the situation is that ud is observation noise of u like wheel odometry.
Yes, that is what I mean. Thank you so mush for fixing the bug and cleaning up it!!
I will merge it.

PR #148, opened 2018-12-18
fixed steer() in RRTstar

luym11:
Please check the [paper](https://core.ac.uk/download/pdf/18173530.pdf), 'Steer()' function in **Algorithm 6** is $steer(x, y) = argmin_{z \in B_{x, \eta}} \Vert z-y \Vert$ and x can be within that ball. Also the cost and parent will be determined by following procedures.


AtsushiSakai:
@luym11 Thank you!!

PR #155, opened 2019-01-14
Fixed a bug of calculating parameters in pure pursuit module

tks10:
I suppose this should be `cy` not `cx`.
I found it because this module didn't work correctly when the path was horizontal.


AtsushiSakai:
@tks10 Thank you for your PR. You are right. I will merge it.

PR #178, opened 2019-03-09
fixed bug of ekf_slam.py

Taka-Kazu:
When I run this script, an error occured.
```
$python3 ekf_slam.py
ekf_slam.py start!!
New LM
Traceback (most recent call last):
  File "ekf_slam.py", line 266, in <module>
    main()
  File "ekf_slam.py", line 234, in main
    xEst, PEst = ekf_slam(xEst, PEst, ud, z)
  File "ekf_slam.py", line 54, in ekf_slam
    xEst = xEst + (K @ y)
TypeError: ufunc 'matmul' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

I found that `xEst[2]`(l. 173) is a 1d-array, so I modified it to be scalar.


AtsushiSakai:
@Taka-Kazu Oh, you're right. Thank you!!

Issue #186, opened 2019-05-09
Runtime optimization for PurePursuit


Kitsunow:
In the `calc_target_index` function you always calculate the distance between all points on the track. I think this is unnecessary, you could optimize this by calculating it once and remember the last target_index, thus only recalculating it for a short range next time:

```
old_target_index = None

def calculate_distance(state, point_x, point_y):
        distance_x = state.rear_x - point_x
        distance_y = state.rear_y - point_y
        return math.sqrt(distance_x ** 2 + distance_y ** 2)

def calc_target_index(state, cx, cy):
    if old_target_index is None:
        # calculate the distances like before, just do the following at the end:
        old_target_index = ind
    else:
        while True:
                distance_this_index = calculate_distance(state, cx[index], cy[index])
                index = index + 1 if (index + 1) < len(path_x) else 0
                distance_next_index = calculate_distance(state, cx[index], cy[index])
                if distance_this_index < distance_next_index:
                    break
                old_target_index = index
```

This way my code only used  1/10 runtime :)

AtsushiSakai:
@Kitsunow Hi. Thank you for your suggestion!! Can you make a PR for it?

Kitsunow:
I created a PR.

Issue #185, opened 2019-05-09
Taking rear axis instead of center for PurePursuit


Kitsunow:
In your pure pursuit example you're using the center of the car for calculation, but in the tech report from Craig Coulter and in a research paper form Jarrod Snider the rear axle is used.

I found significant better behaviour when implementing it with this variant in the physical world, thus asking you if you could change this in your example.

Instead of using `state.x` directly you could modify the state by adding the rear axle as 
```
offset_x = (L / 2) * math.cos(state.yaw)
offset_y = (L / 2) * math.sin(state.yaw)
state.rear_x = state.x - offset_x
state.rear_y = state.y - offset_y
```
and use `state.rear_x` and `state.rear_y` instead of the center now.

AtsushiSakai:
@Kitsunow Hi. Thank you for your suggestion!! Can you make a PR for it?

Kitsunow:
I created a PR.

PR #187, opened 2019-05-09
Optimize PurePursuit, use right axle

Kitsunow:
Two changes made:

- calculate distance between car and all points on path only once, then recalculate based on the old nearest point index and short range until the distance increases

- use rear axle instead of car center


AtsushiSakai:
@Kitsunow Thank you for your great PR!!. I merged it.

AtsushiSakai:
close #186  #185

PR #190, opened 2019-05-22
Fixed the sampling function from PRM

AfroDisco:
When creating a simple square of obstacles, the sampled points are not inside of it, if it is centered around (0, 0)


AtsushiSakai:
@AfroDisco Thank you so much. You are right!!.

PR #192, opened 2019-05-24
Update obstacle map to use appropriate base

jennib3:
Obstacle map only works if minx and miny are 0,0.

### DIFF
--- a/PathPlanning/AStar/a_star.py
+++ b/PathPlanning/AStar/a_star.py
@@ -158,7 +158,7 @@ def calc_obstacle_map(ox, oy, reso, vr):
             for iox, ioy in zip(ox, oy):
                 d = math.sqrt((iox - x)**2 + (ioy - y)**2)
                 if d <= vr / reso:
-                    obmap[ix][iy] = True
+                    obmap[ix+minx][iy+miny] = True
                     break
 
     return obmap, minx, miny, maxx, maxy, xwidth, ywidth


AtsushiSakai:
@jennib3 Hi. Thank you for your PR.

PathPlanning/AStar/a_star.py
```diff
@@ -158,7 +158,7 @@ def calc_obstacle_map(ox, oy, reso, vr):
             for iox, ioy in zip(ox, oy):
                 d = math.sqrt((iox - x)**2 + (ioy - y)**2)
                 if d <= vr / reso:
-                    obmap[ix][iy] = True
+                    obmap[ix+minx][iy+miny] = True
```
AtsushiSakai: I think it is wrong. Because ix is an index of grid (it is from 0 to width or height), so you don't need to add minx or miny.

jennib3: I loaded my own obstacle map and the relative positions of the plotted boundaries and start / goal positions lined up, but the A* algorithm did not avoid the obstacles. Only when I shifted them into the the x/y space did the A* algorithm appropriately mark them.

AtsushiSakai: Oh, ok. as you can see 
https://github.com/AtsushiSakai/PythonRobotics/blob/0be3ea3cf43e8109a74d7fbd60ac380591b989ee/PathPlanning/AStar/a_star.py#L153 
ix is from 0 to xwidth
    for ix in range(xwidth):
so, if minx is a minus value, the index ix+minx become also minus.
This will be error because ix+minx must be positive.
So, it might be another issue of it. 
What do you think?

AtsushiSakai:
@jennib3 Hi. I found the bug which you said. I fix the code by myself. Please check recent master code.Thank you for your report : )

PR #194, opened 2019-05-29
Code live lock when no path is found

AfroDisco:
Added a simple boolean variable to track the state of the algorithm and return something empty when no path is found.
Added and fixed (partially) the documentation


AtsushiSakai:
@AfroDisco Thank you for your PR!!

PR #204, opened 2019-06-18
Once find the path, stop RRT*

Flandoe:
I think It's not necessary to generate the path until the maxIter,  so I modified the program. Now it will stop once it finds a path to the goal.

### DIFF
--- a/PathPlanning/RRTstar/rrt_star.py
+++ b/PathPlanning/RRTstar/rrt_star.py
@@ -63,12 +63,14 @@ class RRT():
             if animation and i % 5 == 0:
                 self.DrawGraph(rnd)
 
-        # generate coruse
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@Flandoe Hi. Thank you for your PR. I think RRT* can improve the path during sampling. So, I think it is OK to keep sampling until maxiter. What do you think?

Flandoe:
@AtsushiSakai I get you. Sometimes trying to find a optimal path is prior, but I think it may be better to find a path as soon as possible because we need use another method to smooth and change the path to meet our conditions in most cases.  It's up to you (๑╹◡╹).

AtsushiSakai:
@Flandoe OK. I see. I will merge it, and I add a control flag for it. Thank you!!

PR #206, opened 2019-06-27
patch the yaw=0 direction while expanding

zhkmxx9302013:
A little change but looks more reasonable.^_^
I patched the yaw=0  when expanding the current node to get neighbors.

[Before]
image

[After]
image


AtsushiSakai:
@zhkmxx9302013 Good catch!! Thank you!!

PR #210, opened 2019-07-08
fixed path

Jumpei-Arima:


AtsushiSakai:
@Jumpei-Arima Good catch!! Thank you.

PR #222, opened 2019-09-07
Update motion_model.py

zhkmxx9302013:
fix the problem of the unsafe type in the `scipy.interpolate`
image


AtsushiSakai:
@zhkmxx9302013 Thank you!!

PR #225, opened 2019-10-01
cvxpy link fixed

sumejko92:
While going through the documentation I noticed that the link the latest version of CVXPY was broken.


AtsushiSakai:
@sumejko92 Oh great!! Thank you so much!!.

PR #228, opened 2019-10-03
convert 1D array to scalar for np.matmul operation

naveenkul:
thrust variable is currently a 1D numpy array so the matmul function throws an error when executed, converting thrust to a scalar before the matmul operation solves the issue.


AtsushiSakai:
@naveenkul Oh great!! Thank you.

PR #239, opened 2019-10-12
Optimize DWA obstacle cost calculation

goktug97:
I've tried to use the algorithm with a grid map that generated with LIDAR data but the algorithm was too slow due to the increased number of obstacles. I've vectorized the calculation and now it is fast to use with a high number of obstacles.


AtsushiSakai:
@goktug97 Great!!. Thank you.

goktug97:
@AtsushiSakai will do, currently, I'm using your matplotlib simulations

PR #241, opened 2019-10-17
 fix #233 improve check segment intersection with obstacle

### LINKED ISSUE #233: Informed RRT* ignores obstacles
@AtsushiSakai it seems informed RRT* suffers from the similar/same issues. 

expandDis = 5, maxIter = 100

path_0

path_1

path_2

I get the same issues with a smaller expandDis...

FranciscoMoretti:
The previous method only worked because the step that checked collision
from a point of the line to obstacles was the same as the minimum
obstacle radius. If the obstacle radius is very small a great amount of
steps would be required.
Thus It's better to check the distance from the segment to the obstacles
directly and compare with obstacle radius
Now there is no need to have two check functions.

Example with same parameters as the issue:
informed rrt

### DIFF
--- a/PathPlanning/InformedRRTStar/informed_rrt_star.py
+++ b/PathPlanning/InformedRRTStar/informed_rrt_star.py
@@ -85,13 +85,13 @@ class InformedRRTStar:
 
                 if self.is_near_goal(newNode):
                     if self.check_segment_collision(newNode.x, newNode.y, self.goal.x , self.goal.y):
-                    solutionSet.add(newNode)
-                    lastIndex = len(self.node_list) - 1
-                    tempPath = self.get_final_course(lastIndex)
-                    tempPathLen = self.get_path_len(tempPath)
-                    if tempPathLen < pathLen:
-                        path = tempPath
-                        cBest = tempPathLen
+                        solutionSet.add(newNode)
+                        lastIndex = len(self.node_list) - 1
+                        tempPath = self.get_final_course(lastIndex)
+                        tempPathLen = self.get_path_len(tempPath)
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@FranciscoMoretti Thank you for your help!!

AtsushiSakai:
#233

AtsushiSakai:
@alex-ira Could you please check this PR fix your issue #233

alex-ira:
@AtsushiSakai Sorry for late reply - been busy lately. I though it did but then got this (with expandDis = 7 - not sure if it is due to not using path_resolution in checking for collisions):

issue

Generally, I get poor performance and many times the path cannot be find (using "reasonable" maxIter). It seems to me that a lot of issues come from checking collisions.

AtsushiSakai:
@alex-ira Thank you for your checking.
@FranciscoMoretti Could you please fix the problem?

AtsushiSakai:
@karanchawla Could you review the code?

FranciscoMoretti:
Sorry for the delay. I've been a little bit busy.
@alex-ira I've found that the problem that you mentioned in the comment was because the solution node there satisfied the condition for "is near goal". The segment from the node to the goal went through obstacles because the script didn't check for collision when a node satisfied that condition. 
I''ve added the missing checking in 178dca3 and it seems to be working fine now.

AtsushiSakai:
In several experiments, it looks working correletry, so I merge it. @FranciscoMoretti Thank you!!

FranciscoMoretti:
@AtsushiSakai Thanks you!!! :)

PR #242, opened 2019-10-23
Fix wrong rotation direction for DWA and Change resampling for the PF

goktug97:
After adding the rectangle obstacle calculation, I've tested the algorithm more to try different settings
and I've seen that sometimes robot hits obstacles, specifically during turns. I've realized, [the yaw angle should be positive instead of negative while rotating the obstacles](https://github.com/AtsushiSakai/PythonRobotics/blob/66fe2f5ef62de5cc01b5ae85f6cb853048b61407/PathPlanning/DynamicWindowApproach/dynamic_window_approach.py#L175). After the change, the robot stops before hitting the obstacles.

**EDIT: About PF Change**: In the book, Probabilistic Robotics and [Introduction to Mobile Robotics Slides ](http://ais.informatik.uni-freiburg.de/teaching/ss19/robotics/slides/09-pf-mcl.pdf) the random variable for the resampling function is chosen between 0 to 1/N only one time but in the code, it is calculated N times. I've changed it so that it matches with the book.

### DIFF
--- a/Localization/particle_filter/particle_filter.py
+++ b/Localization/particle_filter/particle_filter.py
@@ -141,9 +141,8 @@ def re_sampling(px, pw):
     N_eff = 1.0 / (pw.dot(pw.T))[0, 0]  # Effective particle number
     if N_eff < NTh:
         w_cum = np.cumsum(pw)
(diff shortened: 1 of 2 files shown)


goktug97:
Closed to create different pull requests for different algorithms.

PR #245, opened 2019-10-26
change PF resampling function to match with the book

goktug97:
- changed resampling function so that it chooses random number once for the starting point.
- condition check is moved from the inside of the resampling function
to the outside of the function, this way the purpose of the
function is much clear. Instead "resampling" every iteration,
resample if the condition holds.


AtsushiSakai:
@goktug97 Thank you so much again!!. You are totally right. your code looks better. I will merge it.

PR #246, opened 2019-10-26
DWA Fix wrong rotation direction

goktug97:
After adding the rectangle obstacle calculation, I've tested the algorithm more to try different settings
and I've seen that sometimes robot hits obstacles, specifically during turns. I've realized, the yaw angle should be positive instead of negative while rotating the obstacles because of the dot order (`a @ rot` instead of `rot @ a`). After the change, the robot stops before hitting the obstacles.


AtsushiSakai:
@goktug97 Good catch!!.

PR #248, opened 2019-10-27
Fix #247 add cBest update in Informed RRT*

### LINKED ISSUE #247: Informed RRT* incorrectly update cBest for ellipse
**Describe the bug**
In function [informed_rrt_star_search](https://github.com/AtsushiSakai/PythonRobotics/blob/master/PathPlanning/InformedRRTStar/informed_rrt_star.py#L38), `cBest` and `pathLen` seems have same meanings. After finding a new path, `pathLen` is used to check whether `cBest` should be updated, but `pathLen` itself is not being updated. 

The performance of the bug is that the ellipse may increase sometimes when you find a new path.
Also, if you print `cBest` everytime ploting the ellipse, you'll find that `cBest` may increase sometimes. 

**Screenshots**
Here is an example:
There is already a path with length 11.808038783431691.

After a new path with longer length 12.487373607747500 is found, cBest is updated.

JTJL:
Remove pathLen in informed RRT* and instead with cBest updating.

### DIFF
--- a/PathPlanning/InformedRRTStar/informed_rrt_star.py
+++ b/PathPlanning/InformedRRTStar/informed_rrt_star.py
@@ -40,7 +40,6 @@ class InformedRRTStar:
         self.node_list = [self.start]
         # max length we expect to find in our 'informed' sample space, starts as infinite
         cBest = float('inf')
-        pathLen = float('inf')
         solutionSet = set()
         path = None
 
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@JTJL Thank you for your PR!!. 
@karanchawla Could you review the code?

karanchawla:
@atsushi-sakai will go over this tomorrow morning. Thanks, @JTJL for the PR.

AtsushiSakai:
It looks working well, I will merge it. @JTJL Thanks!!

PR #255, opened 2019-11-28
EKF-SLAM: Fix jacobian of observation

m-dobler:
I believe the partial derivative of an observation with respect to the angle should be equal to `-1`, yet in the code it's currently `-1/q`.

Here is a source I could find on the topic: 
http://ais.informatik.uni-freiburg.de/teaching/ws12/mapping/pdf/slam04-ekf-slam.pdf, Slide 33
jacobian_observation

Would greatly appreciate it if someone would take a second look!


AtsushiSakai:
@m-dobler Good catch!! I think you are right. Thank you!!

PR #258, opened 2019-12-06
EKF-SLAM: Fix prediction step of motion

m-dobler:
Fixes the computation of the predicted covariances after a motion command, by changing elementwise multiplication to matrix multiplication.

### DIFF
--- a/SLAM/EKFSLAM/ekf_slam.py
+++ b/SLAM/EKFSLAM/ekf_slam.py
@@ -31,7 +31,7 @@ def ekf_slam(xEst, PEst, u, z):
     S = STATE_SIZE
     xEst[0:S] = motion_model(xEst[0:S], u)
     G, Fx = jacob_motion(xEst[0:S], u)
-    PEst[0:S, 0:S] = G.T * PEst[0:S, 0:S] * G + Fx.T * Cx * Fx
+    PEst[0:S, 0:S] = G.T @ PEst[0:S, 0:S] @ G + Fx.T @ Cx @ Fx
     initP = np.eye(2)
 
     # Update
@@ -119,7 +119,7 @@ def jacob_motion(x, u):
                    [0.0, 0.0, DT * u[0] * math.cos(x[2, 0])],
                    [0.0, 0.0, 0.0]])
 
-    G = np.eye(STATE_SIZE) + Fx.T * jF * Fx
+    G = np.eye(STATE_SIZE) + Fx.T @ jF @ Fx
 
     return G, Fx,
 
--- a/SLAM/EKFSLAM/ekf_slam.ipynb
+++ b/SLAM/EKFSLAM/ekf_slam.ipynb
@@ -309,7 +309,7 @@
     "    G, Fx = jacob_motion(xEst[0:S], u)\n",
     "    # Fx is an an identity matrix of size (STATE_SIZE)\n",
     "    # sigma = G*sigma*G.T + Noise\n",
-    "    PEst[0:S, 0:S] = G.T * PEst[0:S, 0:S] * G + Fx.T * Cx * Fx\n",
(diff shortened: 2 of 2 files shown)


SLAM/EKFSLAM/ekf_slam.py
```diff
@@ -30,8 +30,8 @@ def ekf_slam(xEst, PEst, u, z):
     # Predict
     S = STATE_SIZE
     xEst[0:S] = motion_model(xEst[0:S], u)
-    G, Fx = jacob_motion(xEst[0:S], u)
-    PEst[0:S, 0:S] = G.T * PEst[0:S, 0:S] * G + Fx.T * Cx * Fx
+    G = jacob_motion(xEst[0:S], u)
+    PEst[0:S, 0:S] = G.T @ PEst[0:S, 0:S] @ G + Cx
```
m-dobler: Changed the operator to matrix multiplication instead of elementwise

SLAM/EKFSLAM/ekf_slam.py
```diff
@@ -112,16 +108,12 @@ def calc_n_lm(x):
 
 
 def jacob_motion(x, u):
-    Fx = np.hstack((np.eye(STATE_SIZE), np.zeros(
-        (STATE_SIZE, LM_SIZE * calc_n_lm(x)))))
```
m-dobler: This function is only called with `x` equal to the robot pose, so `calc_n_lm(x)` is 0 and `Fx` is the identity matrix. This change makes the function simpler but removes the capability to pass state including land marks to the function, which is not used.

SLAM/EKFSLAM/ekf_slam.py
AtsushiSakai: Oh. Thank is a bug. Thanks.

SLAM/EKFSLAM/ekf_slam.py
```diff
@@ -94,15 +94,11 @@ def observation(xTrue, xd, u, RFID):
...
 def motion_model(x, u):
-    F = np.array([[1.0, 0, 0],
-                  [0, 1.0, 0],
-                  [0, 0, 1.0]])
-
     B = np.array([[DT * math.cos(x[2, 0]), 0],
                   [DT * math.sin(x[2, 0]), 0],
                   [0.0, DT]])
 
-    x = (F @ x) + (B @ u)
```
AtsushiSakai: Could you please keep (F@x)?. Because, this repo's aim is for learning algorithm. So the code should be same as algorithm. In this simulation case, it is possible to simplify the code, but in other simulation, it might be impossible.

SLAM/EKFSLAM/ekf_slam.py
AtsushiSakai: I would like to keep the code same as before, the reason is same as upper my answer.

SLAM/EKFSLAM/ekf_slam.py
```diff
@@ -112,16 +108,12 @@ def calc_n_lm(x):
...
 
 def jacob_motion(x, u):
-    Fx = np.hstack((np.eye(STATE_SIZE), np.zeros(
-        (STATE_SIZE, LM_SIZE * calc_n_lm(x)))))
-
     jF = np.array([[0.0, 0.0, -DT * u[0] * math.sin(x[2, 0])],
                    [0.0, 0.0, DT * u[0] * math.cos(x[2, 0])],
                    [0.0, 0.0, 0.0]])
 
-    G = np.eye(STATE_SIZE) + Fx.T * jF * Fx
```
AtsushiSakai: I would like to keep the Fx.T and Fx, the reason is same as upper my answer.

SLAM/EKFSLAM/ekf_slam.py
m-dobler: I see your point, thanks for explaining. I undid those commits now and only kept the changes from elementwise to matrix multiplication.

PR #259, opened 2019-12-07
Close on key

goktug97:
If the plot is dynamic such that the plt.pause is used to show the plot, it doesn't let you use the terminal if you are using a window manager such as i3. You have to spam Ctrl+C while clicking to the terminal to close early. Added 
```python
plt.gcf().canvas.mpl_connect('key_release_event',
        lambda event: [exit(0) if event.key == 'escape' else None])
```
statement to every script that uses plt.pause to create dynamic plots, if the plots are static and drawn only once, the statements is not added. Now the scripts can be closed by pressing <kbd>ESC</kbd>.


AtsushiSakai:
@goktug97 Thank you for interesting PR. Can I ask a question?. 
> It doesn't let you use the terminal if you are using a window manager such as i3
Do you want to use terminal during simulation? I don't know whether it is useful or not because I am not a user of i3.

goktug97:
> @goktug97 Thank you for interesting PR. Can I ask a question?.
> 
> > It doesn't let you use the terminal if you are using a window manager such as i3
> 
> Do you want to use terminal during simulation? I don't know whether it is useful or not because I am not a user of i3.

I don't need to use the terminal but if I want to use one, I can change workspace. The problem is, if I want to close the simulation it is al little bit hard and annoying because the matplotlib plot changes the focus so I need to click to terminal and press Ctrl + C at the same time.

AtsushiSakai:
@goktug97 OK. I got it. I would like to merge this PR, but I have a request. Please check the review.

AerialNavigation/drone_3d_trajectory_following/Quadrotor.py
```diff
@@ -24,6 +23,9 @@ def __init__(self, x=0, y=0, z=0, roll=0, pitch=0, yaw=0, size=0.25, show_animat
         if self.show_animation:
             plt.ion()
             fig = plt.figure()
+            fig.canvas.mpl_connect('key_release_event',
```
AtsushiSakai: Could you please add a single line comment for all mpl_connect function like:
    # for stopping simulation with the esc key.

goktug97:
```python
from pathlib import Path

string = 'mpl_connect'
for filename in Path('PythonRobotics').rglob('*.py'):
    f = open(filename, 'r')
    with open(filename, 'r') as f:
        lines = f.readlines()
    numbers = []
    for i, line in enumerate(lines):
        if string in line:
            numbers.append(i)
    for number in reversed(numbers):
        ws = len(lines[number]) - len(lines[number].lstrip(' '))
        lines.insert(number, (ws * ' ') + '# for stopping simulation with the esc key.\n')
    with open(filename, 'w') as f:
        f.writelines(lines)
```
Kind of back fired but done.

AtsushiSakai:
@goktug97 Thank you!!

PR #261, opened 2019-12-07
Simplified calc_target_index function in stanley_controller.py

Gjacquenot:
Used:

- np.hypot to replace `np.sqrt(x**2 + y**2)`
- np.argmin to find location of a minimum in a numpy array


AtsushiSakai:
@Gjacquenot Looks better. Thank you!!

PR #262, opened 2019-12-07
Replaced sqrt(x**2 + y**2) with hypot(x,y)

Gjacquenot:
I made one commit per file. These can be squashed if wanted...


AtsushiSakai:
@Gjacquenot Looks good. Thanks!!

PR #266, opened 2020-01-03
New way to calc covariance for particle filter

FXCarl:
the err ellipse looks strange in original version, so build a new one and looks ok


AtsushiSakai:
@FXCarl Thank you for your PR. I fixed the code based on the paper https://arxiv.org/pdf/1801.07000.pdf 
If there is still any problem in the code, please make another issue or PR.

PR #275, opened 2020-01-24
Fixed float to run PF

ignaciotb:
Fixed ZeroDivisionError when starting the PF

### DIFF
--- a/Localization/particle_filter/particle_filter.py
+++ b/Localization/particle_filter/particle_filter.py
@@ -140,7 +140,7 @@ def re_sampling(px, pw):
     """
 
     w_cum = np.cumsum(pw)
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@ignaciotb Hi. Thank you for your PR. I have a question. When does the problem happen?  In my environment, ZeroDivisionError have never happen..

ignaciotb:
Hi @AtsushiSakai. The problem might rise because of my python version (2.7). I understand it's not supported anymore, but since it's a quick and compatible fix I thought it'd be worth sharing it

AtsushiSakai:
@ignaciotb I'm sorry I don't want to change code for Python2.7 even it is a small change. I'm sorry.

PR #280, opened 2020-02-01
[Kalman Basics] Align some subscripts on the text to follow the code

yoneken:
Some subscripts on the text are different from the actual code. It possibly confuses beginners.


AtsushiSakai:
Ref #280

PR #291, opened 2020-02-10
fix RRT rewire

### LINKED ISSUE #285: RRT star rewire() not work
In RRTStar class, 

    def rewire(self, new_node, near_inds):
        for i in near_inds:
            near_node = self.node_list[i]
            edge_node = self.steer(new_node, near_node)
            if not edge_node:
                continue
            edge_node.cost = self.calc_new_cost(new_node, near_node)

            no_collision = self.check_collision(edge_node, self.obstacle_list)
            improved_cost = near_node.cost > edge_node.cost

            if no_collision and improved_cost:
                near_node = edge_node                    **near_node now points to a new node not in the tree**
                near_node.parent = new_node          **The parent of nodes in the tree does not actually change **
                self.propagate_cost_to_leaves(new_node)

Comment 'near_node = edge_node' can fix this bug. But I think you should check collision between near_node and new_node.

sldai:
fix #285

### DIFF
--- a/PathPlanning/RRTStar/rrt_star.py
+++ b/PathPlanning/RRTStar/rrt_star.py
@@ -34,7 +34,7 @@ class RRTStar(RRT):
             self.cost = 0.0
 
     def __init__(self, start, goal, obstacle_list, rand_area,
-                 expand_dis=3.0,
+                 expand_dis=30.0,
                  path_resolution=1.0,
                  goal_sample_rate=20,
                  max_iter=300,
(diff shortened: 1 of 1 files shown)


PathPlanning/RRTStar/rrt_star.py
```diff
@@ -141,6 +141,9 @@ def search_best_goal_node(self):
     def find_near_nodes(self, new_node):
         nnode = len(self.node_list) + 1
         r = self.connect_circle_dist * math.sqrt((math.log(nnode) / nnode))
+        # if expand_dist exists, search vertices in a range no more than expand_dist
+        if hasattr(self, 'expand_dis'): 
+            r = min(r, self.expand_dis)
```
AtsushiSakai: Is it from original rrt star paper? what is it for?

PathPlanning/RRTStar/rrt_star.py
```diff
@@ -158,8 +161,7 @@ def rewire(self, new_node, near_inds):
             improved_cost = near_node.cost > edge_node.cost
 
             if no_collision and improved_cost:
-                near_node = edge_node
-                near_node.parent = new_node
+                self.node_list[i] = edge_node
```
AtsushiSakai: oh. you are right. this is a bug.

PathPlanning/RRTStar/rrt_star.py
sldai: Yes, this is from the paper. If you specified expanding distance, you must search the near neighbors in that range, so that neighbors are actually accessible from the new sample. 

But if the expanding distance does not exist, like LQR rrt*. You just search in the specified circle, since it's possible to steer the new sample to neighbors in that range.

AtsushiSakai:
@sldai Thank you!!

PR #305, opened 2020-03-24
EKF-SLAM: Fix parameter of Jacobian of motion model

m-dobler:
I believe the Jacobian of the motion model is computed using a wrong parameter.


SLAM/EKFSLAM/ekf_slam.py
```diff
@@ -29,8 +29,8 @@
 def ekf_slam(xEst, PEst, u, z):
     # Predict
     S = STATE_SIZE
-    xEst[0:S] = motion_model(xEst[0:S], u)
     G, Fx = jacob_motion(xEst[0:S], u)
+    xEst[0:S] = motion_model(xEst[0:S], u)
```
m-dobler: Switched the order of the two statements, because the `jacob_motion` should use the expected robot pose **prior** to executing the motion command. 

Corresponding reference from the Probabilistic robotics book: 

Bildschirmfoto 2020-03-24 um 13 06 51

AtsushiSakai: Oh good catch!!

PR #306, opened 2020-04-09
fix compute jacobian of F in ekf_estimation

ndinhtuan:
Following to Eq. (3.51) in Probabilistic Robotic book, G is depends on u_t and \mu_{t-1}, so I think in this code, jacob_F is depend on previous state instead of current predict state


AtsushiSakai:
@ndinhtuan Good catch!! Thank you.

PR #311, opened 2020-04-10
Spanning Tree Search

Guilyx:
Implement Spanning Tree Search (slightly different than DFS but same same) algorithm.
And minor fix in AStar.


Guilyx:
Duplicate of DFS.

Issue #325, opened 2020-05-03
Bug Report : Hybrid A*


guilyx:
Hey @AtsushiSakai, stumbled upon this when I tried the HybridAStar script :

**Describe the bug**
```
$ python hybrid_a_star.py
Start Hybrid A* planning
Traceback (most recent call last):
  File "/home/erwin/.local/lib/python3.7/site-packages/numpy/core/function_base.py", line 117, in linspace
    num = operator.index(num)
TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "hybrid_a_star.py", line 466, in <module>
    main()
  File "hybrid_a_star.py", line 447, in main
    start, goal, ox, oy, XY_GRID_RESOLUTION, YAW_GRID_RESOLUTION)
  File "hybrid_a_star.py", line 341, in hybrid_a_star_planning
    for neighbor in get_neighbors(current, config, ox, oy, obkdtree):
  File "hybrid_a_star.py", line 140, in get_neighbors
    for steer, d in calc_motion_inputs():
  File "hybrid_a_star.py", line 133, in calc_motion_inputs
    for steer in np.concatenate((np.linspace(-MAX_STEER, MAX_STEER, N_STEER),[0.0])):
  File "<__array_function__ internals>", line 6, in linspace
  File "/home/erwin/.local/lib/python3.7/site-packages/numpy/core/function_base.py", line 121, in linspace
    .format(type(num)))
TypeError: object of type <class 'float'> cannot be safely interpreted as an integer.
```

**Expected behavior**
Running Hybrid AStar for the car model...

**Screenshots**
cf current behavior for traceback

**Desktop:**
```shell
$ python --version
Python 3.7.5
```
```shell
$ pip show numpy
Name: numpy
Version: 1.18.4
```
```shell
$ cat /etc/*-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.4 LTS"
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"
```

**Fix:**
N_STEER should be an integer in np.linespace(x, y, N_STEER) and it's a float. Converting it to integer should fix it

AtsushiSakai:
@guilyx Can you make a PR to fix this?

PR #327, opened 2020-05-05
Fix Hybrid A*

guilyx:
Fix issue #325

### DIFF
--- a/PathPlanning/HybridAStar/hybrid_a_star.py
+++ b/PathPlanning/HybridAStar/hybrid_a_star.py
@@ -15,16 +15,19 @@ import sys
 sys.path.append("../ReedsSheppPath/")
 try:
     from a_star_heuristic import dp_planning  # , calc_obstacle_map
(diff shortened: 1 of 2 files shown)


AtsushiSakai:
Thank you!! Is it possible to add a test? It is ok just adding a new test code in the tests dir.
a_star test code might help: https://github.com/AtsushiSakai/PythonRobotics/blob/master/tests/test_a_star.py

guilyx:
@lentychang indeed !

guilyx:
Added the tests and changed `N_STEER` according to @lentychang's suggestion.
@AtsushiSakai as `E722 do not use bare 'except` and all the tests uses bare excepts there might be something to do.

AtsushiSakai:
@guilyx @lentychang Thank you guys!!. I will merge it. 

>  all the tests uses bare excepts there might be something to do.
Oh. Yes. these codes should do it. That's why I'm using incremental code check : ).

PR #345, opened 2020-07-01
Potential field - potential range and ocillations

adi-vardi-enway:
2 changes:
1.  Potential field cover both start and goal - previously potential field was only defined at a certain area around any obstacles. So if goal/start are outside the area, the potential field is undefined, leading to the planner not converging. This fix extends the potential filed to include the start and goal if they are outside the obstacles area.
2. Add simple oscillations detection with horizon of the last 3 previous poses.

### DIFF
--- a/PathPlanning/PotentialFieldPlanning/potential_field_planning.py
+++ b/PathPlanning/PotentialFieldPlanning/potential_field_planning.py
@@ -20,11 +20,11 @@ AREA_WIDTH = 30.0  # potential area width [m]
 show_animation = True
 
 
-def calc_potential_field(gx, gy, ox, oy, reso, rr):
-    minx = min(ox) - AREA_WIDTH / 2.0
-    miny = min(oy) - AREA_WIDTH / 2.0
-    maxx = max(ox) + AREA_WIDTH / 2.0
-    maxy = max(oy) + AREA_WIDTH / 2.0
+def calc_potential_field(gx, gy, ox, oy, reso, rr, sx, sy):
+    minx = min(min(ox),sx,gx) - AREA_WIDTH / 2.0
+    miny = min(min(oy),sy,gy) - AREA_WIDTH / 2.0
+    maxx = max(max(ox),sx,gx) + AREA_WIDTH / 2.0
+    maxy = max(max(oy),sy,gy) + AREA_WIDTH / 2.0
     xw = int(round((maxx - minx) / reso))
     yw = int(round((maxy - miny) / reso))
 
@@ -87,7 +87,7 @@ def get_motion_model():
 def potential_field_planning(sx, sy, gx, gy, ox, oy, reso, rr):
 
     # calc potential field
-    pmap, minx, miny = calc_potential_field(gx, gy, ox, oy, reso, rr)
+    pmap, minx, miny = calc_potential_field(gx, gy, ox, oy, reso, rr, sx, sy)
 
     # search path
     d = np.hypot(sx - gx, sy - gy)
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
@adi-vardi-1 Thank you for great PR!!. I have some review comments.

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
```diff
@@ -106,14 +106,17 @@ def potential_field_planning(sx, sy, gx, gy, ox, oy, reso, rr):
 
     rx, ry = [sx], [sy]
     motion = get_motion_model()
+    previous_id = [(None, None)] * 3
```
AtsushiSakai, line 109: Maybe the variable name should be "xy_history" or something like that. And number 3 should be global parameters like "AREA_WIDTH".

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
```diff
@@ -20,11 +20,11 @@
 show_animation = True
 
 
-def calc_potential_field(gx, gy, ox, oy, reso, rr):
-    minx = min(ox) - AREA_WIDTH / 2.0
-    miny = min(oy) - AREA_WIDTH / 2.0
-    maxx = max(ox) + AREA_WIDTH / 2.0
-    maxy = max(oy) + AREA_WIDTH / 2.0
+def calc_potential_field(gx, gy, ox, oy, reso, rr, sx, sy):
```
AtsushiSakai, line 23: good catch. you are right. thank you.

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
```diff
@@ -128,6 +131,19 @@ def potential_field_planning(sx, sy, gx, gy, ox, oy, reso, rr):
         rx.append(xp)
         ry.append(yp)
 
+        if ((None, None) not in previous_id and
```
AtsushiSakai, line 134: I think this if evaluation should be a boolean function like "is_oscilication_detected" for readability.

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
```diff
@@ -128,6 +131,19 @@ def potential_field_planning(sx, sy, gx, gy, ox, oy, reso, rr):
...
+                (previous_id[0] == previous_id[1]
+                 or previous_id[1] == previous_id[2]
+                 or previous_id[0] == previous_id[2])):
+            print("Oscillation detected!!!")
+            print(previous_id)
+            break
+
+        # roll previous
+        previous_id[0] = previous_id[1]
```
AtsushiSakai, line 143: This code can be used for only when horizon is 3. So, I think "deque" should be used here. https://docs.python.org/3/library/collections.html#collections.deque

adi-vardi-enway: done

adi-vardi-1: done

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
adi-vardi-1: good point, done

PathPlanning/PotentialFieldPlanning/potential_field_planning.py
adi-vardi-1: added

adi-vardi-enway:
great :)

PR #375, opened 2020-08-10
support no obstacle in RRT*

mahyaret:
since `last_index` is 0 when there is no obstacle, the planning will return no path found. Checking for `last_index` being `None` will resolve this issue.

### DIFF
--- a/PathPlanning/RRTStar/rrt_star.py
+++ b/PathPlanning/RRTStar/rrt_star.py
@@ -87,7 +87,7 @@ class RRTStar(RRT):
         print("reached max iteration")
 
         last_index = self.search_best_goal_node()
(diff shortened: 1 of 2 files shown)


AtsushiSakai:
Hi. Thank you for great PR!!. I like your idea, it should be merged. Before that, could you please add a test for no obstacle here as `test_no_obstacle`? https://github.com/AtsushiSakai/PythonRobotics/blob/765f752165d157b3342e9175380befc3b92e3d89/tests/test_rrt_star.py#L22

mahyaret:
Thank you for your great repo! I've added a test as requested.

AtsushiSakai:
LGTM. Thanks!!

PR #381, opened 2020-08-23
Changed beta variable to have a nicer semantic.

samlaf:
controller was of the form
w = Kp_alpha * alpha + Kp_beta * beta
and beta was theta_diff - alpha... which gives
w = Kp_alpha * alpha + Kp_beta (theta_diff - alpha) = (Kp_alpha - Kp_beta)*alpha + Kp_beta*theta_diff
So I made Kp_alpha = 18 = 15 - (-3)
and changed beta to just theta_diff = theta_goal - theta.

### DIFF
--- a/PathTracking/move_to_pose/move_to_pose.py
+++ b/PathTracking/move_to_pose/move_to_pose.py
@@ -15,7 +15,7 @@ from random import random
 
 # simulation parameters
 Kp_rho = 9
-Kp_alpha = -18
+Kp_alpha = 18
 Kp_beta = -3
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
hi, thank you for your PR. I have a question. why do you think it is nicer?

samlaf:
My brother made me realize that there might be a geometric interpretation to Beta which I missed, I'm sorry...
It would be nice to have a picture like this in the docs.
IMG_DB56926CD16B-1

PR #385, opened 2020-08-29
DWA_robot_stucked_issue

weicent:
1. add the brake velocity into consideration (the max velocity that robot can brake in front of an obstacle) 
2. provide a possible solution for robot to escape from stucked


AtsushiSakai:
Hi. Thank you for your PR. Could you please fix all CI errors?

Issue #379, opened 2020-08-19
Dynamic window approach robot stucked


weicent:
**Describe the bug**
Hello Atsushi

Thanks for creating this wonderful repository.

I'm working on dynamic window approch recently, and I found a bug for the code you uploaded.

when I using below obstacle and goal position, the robot stucked in a certain angle.

    ob =-1* np.array([[-1, -1],
                   [0, 2],
                   [2, 6],
                   [2, 8],
                   [3, 9.27],
                   [3.79, 9.39],
                   [7.25, 8.97],
                   [7.0, 2.0],
                   [3.0, 4.0],
                   [6.0, 5.0],
                   [3.5, 5.8],
                   [6.0, 9.0],
                   [8.8, 9.0],
                   [5.0, 9.0],
                   [7.5, 3.0],
                   [9.0, 8.0],
                   [5.8, 4.4],
                   [12.0, 12.0],
                   [3.0, 2.0],
                   [13.0, 13.0]
                   ])

goal = [-5, -7]

I studied in this case and found the robot change its heading to reduce the angle difference with the goal, then it reaching the best angle difference (which means robot is derectly heading to the goal), so it will not change the angle, however, in this case, there is just an obstacle in front of robot, which lead to a best velocity equal to zero.

Under this circumstance, the robot will get stucked into endless loop (angle difference to the goal = 0 rad, velocity = 0 m/s).

I finally solved this by checking the robot status when running the while loop:

```
#below is not python code, just for illustration
if robot speed<0.1 (or some tiny value close to 0):
      if robot_angle_difference_with_the_goal<0.01 (or some angle close to 0 rad):
             best_u = [best_u[0], best_u[1] + config.angular_acceleration* config.dt]
```

this two if-case will continually check the robot status and ensure the robot will not stucked in that case, but this might not be an efficient method, i will be very appreciate if you can share any better solution.

By the way, on the other hand, it seems taht you didn't consider the brake distance for the dynamic window, (i'm not sure on this point, ignore me if i'm worng).

when generate speed window, the maximum speed allowed for stop in distance S is Vmax=sqrt(2*S*a)

so if you consider this brake velocity, the method for dw() should like below (i appologize if i'm worng):

```
def dynamic_window(x, traj):
    # max speed it can reach
    vmin = x[3] - para.vv * para.dt
    vmax = x[3] + para.vv * para.dt
    wmin = x[4] - para.ww * para.dt #w for angular velocity
[...]

PR #390, opened 2020-09-01
dwa pr

weicent:
provide a solution for robot to escape from stucked condition 
 Issue description:
[https://github.com/AtsushiSakai/PythonRobotics/issues/379#issue-681892447](url)


AtsushiSakai:
Rerun CI

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
```diff
@@ -232,29 +235,34 @@ def plot_robot(x, y, yaw, config):  # pragma: no cover
         plt.plot([x, out_x], [y, out_y], "-k")
 
 
-def main(gx=10.0, gy=10.0, robot_type=RobotType.circle):
+def main(gx=-5.0, gy=-7.0, robot_type=RobotType.circle):
```
AtsushiSakai, line 238: Could you please change your main function to main2 or something and keep previous main function in order to fix unittest.

AtsushiSakai (changes requested):
Hi. I looks like your code change break CI. Please check my comment to fix the test.

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
weicent: Hi Atsushi, I have tried to change the name of 'main' function into 'test_pr' and keep the original 'main' function unchanged, then the CI test indicate 'xxx was not covered by tests', which means the 'test_pr' function was not used, therefore i deleted the 'test_pr' function and only keep the original main function.

below is the data which cause robot stucked issue in original code (in case you want replace the data and test the code):
goal = [-5, -7]
ob =-1* np.array([[-1, -1],
               [0, 2],
               [2, 6],
               [2, 8],
               [3, 9.27],
               [3.79, 9.39],
               [7.25, 8.97],
               [7.0, 2.0],
               [3.0, 4.0],
               [6.0, 5.0],
               [3.5, 5.8],
               [6.0, 9.0],
               [8.8, 9.0],
               [5.0, 9.0],
               [7.5, 3.0],
               [9.0, 8.0],
               [5.8, 4.4],
               [12.0, 12.0],
               [3.0, 2.0],
               [13.0, 13.0]
               ])

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
```diff
@@ -101,7 +100,6 @@ def calc_dynamic_window(x, config):
           x[4] - config.max_delta_yaw_rate * config.dt,
           x[4] + config.max_delta_yaw_rate * config.dt]
 
-    #  [v_min, v_max, yaw_rate_min, yaw_rate_max]
```
AtsushiSakai, line 104: Could you please put back the comment? I think it is useful.

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
```diff
@@ -152,13 +149,18 @@ def calc_control_and_trajectory(x, dw, config, goal, ob):
                 min_cost = final_cost
                 best_u = [v, y]
                 best_trajectory = trajectory
-
+                if abs(best_u[0]) < 0.001 and abs(x[3]) < 0.001:
```
AtsushiSakai, line 152: Could you please create two class const variable for 0.001 and 0.001 with proper variable name?

AtsushiSakai (changes requested):
I have some comments. PTAL. BTW I have two other comments.
1. Where is braking distance code which you mentioned in the issue?
2. It is ok to add your test code in test file even code cov CI report error, maybe we can fix it.

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
weicent: No problem, the 0.001 is changed to config.robot_stuck_flag_cons.

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
weicent: Sorry for the missing comment, I have made some changes in this part before for the brake velocity limit, however when testing the code, I found the robot keep moves in a very low speed thus delete this change, below is the explanation for the idea:
in order to brake in front of an obstacle, the max speed limitation should not greater than sqrt(2*distance*acceleration), therefore I bring this into consideration when generate dynamic window.
The basic idea is find the distance to the nearest obstacle, then calculate the brake speed limit and compare with the Vd and Vs in the original code to find the dynamic window.
This changes handle the situations that the original code failed, which is in some obstacle maps (the obstacle map and goal I have provided before), the original code drive the circular shape robot just across through the obstacles because the velocity is too high to brake (while the rectangle robots works fine).
However, this change also bring a problem which is I don’t know how to filter the below situation in a simple manner: 

demo

As can be seen from figure, in this situation, the obstacle is not in the way of robot, so we should not consider the brake speed when generate dynamic window, the brake velocity should only be considered when the obstacle is just in front of the robot heading.
My changes in brake velocity part didn’t filter this situation, therefore although it can handle some problems, the speed is significantly limited since the robot is continueously assume it need to brake in front of the nearest obstacle (which actually may not in its way), this makes the robot moves very slow.

Under this consideration, I delete the brake velocity part and modified the to_goal_cost_gain=0.2 and obstacle_cost_gain=2.0, this helps to solve the robot cross obstacle problems and meanwhile keep the robot move in a higher speed.

AtsushiSakai (changes requested):
Thank you for update code. This is the last review, after fixed it I will merge it.

Could you please add a test here for test your stack case?
https://github.com/AtsushiSakai/PythonRobotics/blob/46bdd2afd950046d3f57c181efb0b867c63a2536/tests/test_dynamic_window_approach.py#L22

Parameter should be these which you specified:
goal = [-5, -7]
ob =-1* np.array([[-1, -1],
[0, 2],
[2, 6],
[2, 8],
[3, 9.27],
[3.79, 9.39],
[7.25, 8.97],
[7.0, 2.0],
[3.0, 4.0],
[6.0, 5.0],
[3.5, 5.8],
[6.0, 9.0],
[8.8, 9.0],
[5.0, 9.0],
[7.5, 3.0],
[9.0, 8.0],
[5.8, 4.4],
[12.0, 12.0],
[3.0, 2.0],
[13.0, 13.0]
])

weicent:
@AtsushiSakai thanks for your reply, in order to update the new obstacle array in test_dwa_stuck_case.py, i have also changed ob into a config property rather  than pre-specified parameter in main(), this will not effect the original test_dynamic_window_approach.py file.

tests/test_dwa_stuck_case.py
```diff
@@ -0,0 +1,45 @@
+import os
```
AtsushiSakai, line 1: Could you please just add a test function in test_dynamic_window_approach.py, not adding a new file?

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
```diff
@@ -48,9 +47,10 @@ def __init__(self):
         self.yaw_rate_resolution = 0.1 * math.pi / 180.0  # [rad/s]
         self.dt = 0.1  # [s] Time tick for motion prediction
         self.predict_time = 3.0  # [s]
-        self.to_goal_cost_gain = 0.15
+        self.to_goal_cost_gain = 0.2
```
AtsushiSakai, line 50: Could you please keep previous parameter as default value and change parameter in the test function?

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
```diff
@@ -48,9 +47,10 @@ def __init__(self):
...
         self.speed_cost_gain = 1.0
-        self.obstacle_cost_gain = 1.0
+        self.obstacle_cost_gain = 2.0
```
AtsushiSakai, line 52: Could you please keep previous parameter as default value and change parameter in the test function?

AtsushiSakai (changes requested):
I'm sorry could you please fix some minor points? PTAL.

PathPlanning/DynamicWindowApproach/dynamic_window_approach.py
weicent: thanks for your patient replay, the file is modified as you suggested.
this is really a tough pr, while i have learned a lot from this process, thanks again for all of your kind suggestion!

AtsushiSakai (approved):
LGTM. I will merge it. Thank you so much!!

PR #402, opened 2020-09-14
robustified 2 joint arm example

ajwahab:
two joint arm point control example will no longer fail if target point is out of reach


AtsushiSakai:
@ajwahab Hi. Thank you so much for your PR. Could you please add a test code before review?

ajwahab:
The change is so minor (basic geometry), I didn't think it was worth the time to write test code.  I tested by running the script and clicking beyond the reach of the arm.  Before my change, the arm would disappear from the plot window when the target was outside the arm's maximum radius.  My change just causes the arm's joint angles to saturate for maximum reach.

AtsushiSakai:
OK. I will add a test by myself. I will merge it. Thanks.

Issue #409, opened 2020-09-27
two_joint_arm_to_point_control.py throw TypeError when clicking outside of the figure.


AtsushiSakai:
**Describe the bug**
two_joint_arm_to_point_control.py throws TypeError when clicking outside of the figure.

**Expected behavior**
Even if click outside of the figure, the simulation should keep working, just ignoring the click event. 

**Screenshots**
Error message is

> Traceback (most recent call last):
>   File "/Users/atsushisakai/PycharmProjects/PythonRobotics/ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py", line 128, in <module>
>     main()
>   File "/Users/atsushisakai/PycharmProjects/PythonRobotics/ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py", line 123, in main
>     two_joint_arm()
>   File "/Users/atsushisakai/PycharmProjects/PythonRobotics/ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py", line 61, in two_joint_arm
>     d2goal = np.hypot(wrist[0] - x, wrist[1] - y)
> TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
> 

**Desktop (please complete the following information):**
 - Python version (This repo only supports Python 3.7.x or higher).
 - Each library version

PR #414, opened 2020-10-02
fixed TypeError

raghuramshankar:
#### Reference issue

#409 

Fixed the bug where when you click outside the figure, it throws TypeError. Now it just ignores the outside clicks.

### DIFF
--- a/ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py
+++ b/ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py
@@ -129,8 +129,8 @@ def main():  # pragma: no cover
     fig = plt.figure()
     fig.canvas.mpl_connect("button_press_event", click)
     # for stopping simulation with the esc key.
-    fig.canvas.mpl_connect('key_release_event',
-                           lambda event: [exit(0) if event.key == 'escape' else None])
+    fig.canvas.mpl_connect('key_release_event', lambda event: [
+                           exit(0) if event.key == 'escape' else None])
     two_joint_arm()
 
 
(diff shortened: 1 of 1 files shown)


ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py
```diff
@@ -54,11 +59,17 @@ def two_joint_arm(GOAL_TH=0.0, theta1=0.0, theta2=0.0):
...
+            y = y_prev
 
         wrist = plot_arm(theta1, theta2, x, y)
 
         # check goal
-        d2goal = np.hypot(wrist[0] - x, wrist[1] - y)
+        try:
+            d2goal = np.hypot(wrist[0] - x, wrist[1] - y)
+        except TypeError:
+            pass
```
AtsushiSakai, line 72: This empty except is needed?

AtsushiSakai (changes requested):
Thank you for your PR again. I have a small comment. PTAL.

ArmNavigation/two_joint_arm_to_point_control/two_joint_arm_to_point_control.py
raghuramshankar: Hi, thank you for your review. Yes, if this error is not passed, the same bug appears again as the value of x and y are "NoneType". I feel it is needed,

AtsushiSakai: OK. So could you please use if statement instead of try catch.

AtsushiSakai (approved):
LGTM. Thank you!!

PR #415, opened 2020-10-02
Adding information on several entries to precise which algorithm is used

lfrering:
Several entries in the readme were a bit unclear about what method is used, I just added a bit of information in them.


AtsushiSakai:
I'm sorry. I think that change is not so important...

PR #427, opened 2020-10-28
Using vectorization to speed up computation

maudzung:

### DIFF
--- a/Mapping/rectangle_fitting/rectangle_fitting.py
+++ b/Mapping/rectangle_fitting/rectangle_fitting.py
@@ -69,15 +69,11 @@ class LShapeFitting:
         c1_min = min(c1)
         c2_min = min(c2)
 
-        D1 = [min([np.linalg.norm(c1_max - ic1),
-                   np.linalg.norm(ic1 - c1_min)]) for ic1 in c1]
-        D2 = [min([np.linalg.norm(c2_max - ic2),
-                   np.linalg.norm(ic2 - c2_min)]) for ic2 in c2]
-
-        beta = 0
-        for i, _ in enumerate(D1):
-            d = max(min([D1[i], D2[i]]), self.min_dist_of_closeness_criteria)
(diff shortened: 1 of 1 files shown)


Mapping/rectangle_fitting/rectangle_fitting.py
```diff
@@ -69,15 +69,11 @@ def _calc_closeness_criterion(self, c1, c2):
...
-        D2 = [min([np.linalg.norm(c2_max - ic2),
-                   np.linalg.norm(ic2 - c2_min)]) for ic2 in c2]
-
-        beta = 0
-        for i, _ in enumerate(D1):
-            d = max(min([D1[i], D2[i]]), self.min_dist_of_closeness_criteria)
-            beta += (1.0 / d)
+        # Vectorization
+        D1 = np.minimum(c1_max - c1, c1 - c1_min)
+        D2 = np.minimum(c2_max - c2, c2 - c2_min)
```
AtsushiSakai, line 74: Could you please explain why np.linalg.norm is not needed?

AtsushiSakai (commented):
Thank you for great PR. You code is more beautiful. I have a question about your change. PTAL.

Mapping/rectangle_fitting/rectangle_fitting.py
maudzung: `c1_max - c1`, `c1 - c1_min`, `c2_max - c2`, `c2 - c2_min` are 1D vectors, and all components in the 1D vectors are greater or equal 0. Hence we don't need to compute norm values and don't need to take the absolute value also.

AtsushiSakai: OK. I understood. Thanks.

AtsushiSakai (approved):
LGTM. Thank you. 
(I found this script does not have a test, so I created a issue in #436)

PR #435, opened 2020-11-07
fixing sign in lane center calculation

### LINKED ISSUE #425: State lattice planner confusion regarding xc and yc calcuation
I think the calculation for (xc, yc) is missing a negative.
The general rotation matrix is: 
[cos(theta) -sin(theta)
 sin(theta)  cos(theta)]

```python
def calc_lane_states(l_center, l_heading, l_width, v_width, d, nxy):
    """
    calc lane states
    :param l_center: lane lateral position
    :param l_heading:  lane heading
    :param l_width:  lane width
    :param v_width: vehicle width
    :param d: longitudinal position
    :param nxy: sampling number
    :return: state list
    """
    xc = math.cos(l_heading) * d + math.sin(l_heading) * l_center
    yc = math.sin(l_heading) * d + math.cos(l_heading) * l_center
```

CtfChan:
#### Reference issue

fix #425 

Using proper rotation matrix in xc, yc calculation for lane sampling

#### Additional information

Please let me know if there's other stuff I'm supposed to change. Never contributed to an open repo before.

### DIFF
--- a/PathPlanning/StateLatticePlanner/state_lattice_planner.py
+++ b/PathPlanning/StateLatticePlanner/state_lattice_planner.py
@@ -156,7 +156,7 @@ def calc_lane_states(l_center, l_heading, l_width, v_width, d, nxy):
     :param nxy: sampling number
     :return: state list
     """
-    xc = math.cos(l_heading) * d + math.sin(l_heading) * l_center
+    xc = math.cos(l_heading) * d - math.sin(l_heading) * l_center
     yc = math.sin(l_heading) * d + math.cos(l_heading) * l_center
 
     states = []
@@ -301,7 +301,7 @@ def lane_state_sampling_test1():
     k0 = 0.0
 
     l_center = 10.0
-    l_heading = np.deg2rad(90.0)
+    l_heading = np.deg2rad(0.0)
     l_width = 3.0
     v_width = 1.0
     d = 10
--- a/PathPlanning/StateLatticePlanner/state_lattice_planner.py
+++ b/PathPlanning/StateLatticePlanner/state_lattice_planner.py
@@ -309,6 +309,9 @@ def lane_state_sampling_test1():
     states = calc_lane_states(l_center, l_heading, l_width, v_width, d, nxy)
     result = generate_path(states, k0)
 
+    if show_animation:
+        plt.close("all")
+
     for table in result:
         xc, yc, yawc = motion_model.generate_trajectory(
             table[3], table[4], table[5], k0)


AtsushiSakai:
Thank you for your PR. I checked this script, the` calc_lane_states` is so buggy. 
For example, when I set this configuration,
```
l_center = 10.0
l_heading = np.deg2rad(45.0)
l_width = 3.0
v_width = 1.0
d = 10
nxy = 5 
```
The terminal state should be, x =10, y =10.0, yaw = 45.0[deg], these input values should be global values.
So, the rotation matrix is not needed I think.

The result should be like
SS 2020-11-08 at 14 35 22

Also, I found another bug. The nxy is the number of the path, so the number of the pass should be 5 in this example,
But there are 6 paths... Can you fix it also.

PathPlanning/StateLatticePlanner/state_lattice_planner.py
```diff
@@ -156,7 +156,7 @@ def calc_lane_states(l_center, l_heading, l_width, v_width, d, nxy):
     :param nxy: sampling number
     :return: state list
     """
-    xc = math.cos(l_heading) * d + math.sin(l_heading) * l_center
+    xc = math.cos(l_heading) * d - math.sin(l_heading) * l_center
     yc = math.sin(l_heading) * d + math.cos(l_heading) * l_center
```
AtsushiSakai, line 160: As i said this rotation is not needed. This line just should be
```
xc = d
yc = l_center
```

AtsushiSakai (changes requested):
PTAL

AtsushiSakai:
One more thing, could you please add this code in line 311 to remove animation result?

```
if show_animation:
   plt.close("all")

```

CtfChan:
> Thank you for your PR. I checked this script, the` calc_lane_states` is so buggy.
> For example, when I set this configuration,
> 
> ```
> l_center = 10.0
> l_heading = np.deg2rad(45.0)
> l_width = 3.0
> v_width = 1.0
> d = 10
> nxy = 5 
> ```
> 
> The terminal state should be, x =10, y =10.0, yaw = 45.0[deg], these input values should be global values.
> So, the rotation matrix is not needed I think.
> 
> The result should be like
> SS 2020-11-08 at 14 35 22

Hi Atsushi,

I added the line to close the animations on 311.

I believe using the rotation matrix is required. Below is a link to another implementation of a state lattice planner that also uses the rotation matrix. 

https://github.com/enginBozkurt/MotionPlanner/blob/8f918944fd992b53e0c10b9e0df18f97a02b51b8/MotionPlannerCourse4FinalProject/local_planner.py#L139

Using the input values you gave I don't think the terminal state should be, x =10, y =10.0, yaw = 45.0[deg].  In the diagram below l_heading describes the rotation of frame b wrt to frame a. The centered terminal state (d, l_center) is wrt to frame b. So we need to rotate it back into frame a.



Let me know what you think.


Cheers.

AtsushiSakai:
I think it depends on the input value on frame a or frame b in your fig. I think using frame a is usual for practical use case than using frame b. If we use frame a, we don't need to use a rotation matrix, I think. What do you think?

CtfChan:
Yes I agree. If you use frame A, you won't need frame B. But I think the shape parametrization p_{xx} used in the paper uses a frame A and B.

AtsushiSakai:
Hmm. I think shape parametrization p_{xx} does not depend on the frame.

PR #447, opened 2020-12-21
fix ICP negative dError. Issue #446

### LINKED ISSUE #446: ICP bug with negative delta error
**Describe the bug**
If at certain iteration the ```error``` of ```nearest_neighbor_association``` starts increasing and we result in negative ```dError```, ICP continues iterations until ```dError <= EPS``` or ``` MAX_ITER <= count ``` and the error keeps increasing.

**Expected behavior**
2 possible ways of behavior:
- Simple way. When ```dError``` is negative, exit loop to prevent increasing the ```error```
- Advanced way. Keep in memory state, where ```error``` was minimum. Continue iterations with negative ```dError```. If finally it converges to state with ```error``` less than saved one - keep last state. If not - keep the saved state.

**Desktop**
 - Python version 3.7.7

NobleBumblebee:
#### Reference issue
fix #446

Fixes case of negative ```dError``` in ICP algorithm.

#### Additional information
Implementation:
When ```dError``` is negative (error is rising) exits iteration loop with values of previous iterations with lower error. 
The first iteration is guaranteed by defining ```dError``` and ```preError``` as np.inf before the loop.

### DIFF
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -48,7 +48,7 @@ def icp_matching(previous_points, current_points):
         indexes, error = nearest_neighbor_association(previous_points, current_points)
         Rt, Tt = svd_motion_estimation(previous_points[:, indexes], current_points)
 
-        if dError < 0:  # if negative error prevent matrix H changing, exit loop
+        if dError < 0:  # prevent matrix H changing, exit loop
             print("Converge", preError, dError, count)
             break
 
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -48,7 +48,7 @@ def icp_matching(previous_points, current_points):
         indexes, error = nearest_neighbor_association(previous_points, current_points)
         Rt, Tt = svd_motion_estimation(previous_points[:, indexes], current_points)
 
-        if dError < 0: # if negative error prevent matrix H changing, exit loop
+        if dError < 0:  # if negative error prevent matrix H changing, exit loop
             print("Converge", preError, dError, count)
             break
 
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -37,8 +37,9 @@ def icp_matching(previous_points, current_points):
         if show_animation:  # pragma: no cover
             plt.cla()
             # for stopping simulation with the esc key.
-            plt.gcf().canvas.mpl_connect('key_release_event',
-                                         lambda event: [exit(0) if event.key == 'escape' else None])
+            plt.gcf().canvas.mpl_connect(
+                'key_release_event',
+                lambda event: [exit(0) if event.key == 'escape' else None])
             plt.plot(previous_points[0, :], previous_points[1, :], ".r")
             plt.plot(current_points[0, :], current_points[1, :], ".b")
             plt.plot(0.0, 0.0, "xr")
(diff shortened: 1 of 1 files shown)


AtsushiSakai:
Thank you for you PR. Your improvement make sense for me. I have some minor requests to improve the whole code. PTAL.

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -48,6 +48,10 @@ def icp_matching(previous_points, current_points):
         indexes, error = nearest_neighbor_association(previous_points, current_points)
         Rt, Tt = svd_motion_estimation(previous_points[:, indexes], current_points)
 
+        if dError < 0:  # prevent matrix H changing, exit loop
```
AtsushiSakai, line 51: dError is only updated in line 60. If we check the value in line 64 with "dError <= EPS", we can check the value is negative or not. So, I think this check is not needed here. What do you think?

AtsushiSakai (changes requested):
PTAL. Also, this is not your fault, but could you please fix indention error in line 41?

SLAM/iterative_closest_point/iterative_closest_point.py
NobleBumblebee: I put ```if dError < 0``` before the line 58 that changes the state of matrix H. If the error increases, we want to use matrix H from the previous iteration when error was lower. So breaking before line 58 keeps previous iteration transformation.

NobleBumblebee:
I am not sure, do you want the indention in line 41 to be the same level as "(" or "plt" + tab?

SLAM/iterative_closest_point/iterative_closest_point.py
AtsushiSakai: > we want to use matrix H from the previous iteration
Yes. I know what you want. But, I am not sure your change achieve it.
dError is only updated in line 60, so if dError changed to negative, it always break at line 64 without you change (if dError <= EPS.). 
I think this means previous code works  based on what you want.

AtsushiSakai:
I'm sorry I mean this code
https://github.com/AtsushiSakai/PythonRobotics/blob/581b3cf27ab188a319162dc13dc926503b52cda9/SLAM/iterative_closest_point/iterative_closest_point.py#L40-L41
The second line should be align the top of (`key.... of the first line.

SLAM/iterative_closest_point/iterative_closest_point.py
NobleBumblebee: Yes, you are right. Thank you for watching. My bad, I didn't transfered the code correctly, as I did some more modifications for my project.
1. dError should not be absolute, otherwise we cannot catch negative dError.
2. dError assignment goes before ```if dError < 0``` check.
3. Also changed from Converge to Not Converge as when error is increasing it's the second case.

NobleBumblebee:
The indention we were talking about didn't passed the check, so I suggest split function arguments in two lines as I did in last commit.

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -48,15 +49,20 @@ def icp_matching(previous_points, current_points):
...
+        print("Residual:", error)
+
+        if dError < 0:  # prevent matrix H changing, exit loop
+            print("Not Converge...", preError, dError, count)
+            break
+
+        preError = error
+
         # update current points
         current_points = (Rt @ current_points) + Tt[:, np.newaxis]
```
AtsushiSakai, line 62: Could you please move line 61 and 62 to line 50 after svd_motion_estimation? I think this looks more easy to read the code.

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -48,15 +49,20 @@ def icp_matching(previous_points, current_points):
...
 
         H = update_homogeneous_matrix(H, Rt, Tt)
 
-        dError = abs(preError - error)
-        preError = error
-        print("Residual:", error)
-
         if dError <= EPS:
```
AtsushiSakai, line 66: We can merge these break statements to line 55 like:

```
if 0 <= dError <= EPS:
    print("Converge", error, dError, count)
    break
elif dError < 0:
    print("Not Converge... Error increased. ", preError, dError, count)
    break
elif MAX_ITER <= count:
     print("Not Converge... Reached max iterations", error, dError, count)
     break
```

AtsushiSakai (changes requested):
Thank you for fixing. I think this code works correctly. But I think we can improve the code for readability. PTAL.

SLAM/iterative_closest_point/iterative_closest_point.py
NobleBumblebee: I cannot see how we can merge break statements. Because it's important for ```H = update_homogeneous_matrix``` to be after ```dError < 0``` (as at this break we take H status from previous iteration) and before ```dError <= EPS``` (as we need to update H before possible breaking for these 2 cases).

P.S. I thought it over, it is possible to move up 2 last break statements, but there will always be something between ```dError < 0``` and the others. Moving them up, it is necessary to do ```dError = preError - error``` after them but before ```dError < 0```. To my mind it becomes a mess in algorithm logic, so I suggest to keep code as is.
I share your wish to make things beautiful. But here I don't see a way to make it better.

AtsushiSakai: OK. Make sense. Thanks.

AtsushiSakai (approved):
LGTM. Thank you for great improvement!!.

PR #465, opened 2021-01-25
Add ICP support for 3d point clouds

### LINKED ISSUE #464: Add ICP support for 3d point clouds
**Describe the bug**
In functions `icp_matching()` and `update_homogeneous_matrix()` processing of point clouds is fixed for a 2d case only. The ICP implementation cannot be used for 3d case.

**Expected behavior**
Modify a code to be able to work for both 2d and 3d cases.

**Desktop (please complete the following information):**
- Python version 3.7.7

NobleBumblebee:
#### Reference issue
fix #464

Adds an implementation of code for ICP to work both for 2d and 3d point clouds.

#### Additional information
Make functions `icp_matching()` and `update_homogeneous_matrix()` flexible. Depending on the `current_points.shape[0]` function `svd_motion_estimation()` automatically returns the correct size for `Rt`, `Tt`. Then, in `update_homogeneous_matrix()` depending on their size matrix `H` is formed to be 3x3 or 4x4. 

Test for 3d point cloud is provided. Made the same manner as tests for 2d point cloud.

### DIFF
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -5,6 +5,7 @@ author: Atsushi Sakai (@Atsushi_twi), Göktuğ Karakaşlı
 
 import math
 
+from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 unused import
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -19,8 +20,8 @@ def icp_matching(previous_points, current_points):
     """
     Iterative Closest Point matching
     - input
-    previous_points: 2D points in the previous frame
-    current_points: 2D points in the current frame
+    previous_points: 2D or 3D points in the previous frame
+    current_points: 2D or 3D points in the current frame
     - output
     R: Rotation matrix
     T: Translation vector
@@ -35,15 +36,24 @@ def icp_matching(previous_points, current_points):
         count += 1
 
         if show_animation:  # pragma: no cover
-            plt.cla()
             # for stopping simulation with the esc key.
             plt.gcf().canvas.mpl_connect(
                 'key_release_event',
                 lambda event: [exit(0) if event.key == 'escape' else None])
-            plt.plot(previous_points[0, :], previous_points[1, :], ".r")
-            plt.plot(current_points[0, :], current_points[1, :], ".b")
-            plt.plot(0.0, 0.0, "xr")
-            plt.axis("equal")
+            if previous_points.shape[0] == 3:
+                fig = plt.figure()
+                ax = fig.add_subplot(111, projection='3d')
+                ax.scatter(previous_points[0, :], previous_points[1, :], 
+                            previous_points[2, :], c = "r", marker = ".")
+                ax.scatter(current_points[0, :], current_points[1, :], 
+                           current_points[2, :], c = "b", marker = ".")
+                ax.scatter(0.0, 0.0, 0.0, c = "r", marker = "x")
+            else:
+                plt.cla()
+                plt.plot(previous_points[0, :], previous_points[1, :], ".r")
+                plt.plot(current_points[0, :], current_points[1, :], ".b")
+                plt.plot(0.0, 0.0, "xr")
+                plt.axis("equal")
             plt.pause(0.1)
 
         indexes, error = nearest_neighbor_association(previous_points, current_points)
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -68,24 +68,20 @@ def icp_matching(previous_points, current_points):
             print("Not Converge...", error, dError, count)
             break
 
-    R = np.array(H[0:2, 0:2])
-    T = np.array(H[0:2, 2])
+    R = np.array(H[0:-1, 0:-1])
+    T = np.array(H[0:-1, -1])
 
     return R, T
 
 
 def update_homogeneous_matrix(Hin, R, T):
 
-    H = np.zeros((3, 3))
+    r_size = R.shape[0]
+    H = np.zeros((r_size + 1, r_size + 1))
 
-    H[0, 0] = R[0, 0]
-    H[1, 0] = R[1, 0]
-    H[0, 1] = R[0, 1]
-    H[1, 1] = R[1, 1]
-    H[2, 2] = 1.0
-
-    H[0, 2] = T[0]
-    H[1, 2] = T[1]
+    H[0:r_size, 0:r_size] = R
+    H[0:r_size, r_size] = T
+    H[r_size, r_size] = 1.0
 
     if Hin is None:
         return H
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -148,8 +148,16 @@ def main():
         print("R:", R)
         print("T:", T)
 
-    # test for 3d point set
+
+def main_3d_points():
+    print(__file__ + " start!!")
+
+    # simulation parameters for 3d point set
+    nPoint = 1000
+    fieldLength = 50.0
     motion = [0.5, 2.0, -5, np.deg2rad(-10.0)]  # [x[m],y[m],z[m],roll[deg]]
+
+    nsim = 3  # number of simulation
 
     for _ in range(nsim):
 
@@ -174,3 +182,4 @@ def main():
 
 if __name__ == '__main__':
     main()
+    main_3d_points()
--- a/SLAM/iterative_closest_point/iterative_closest_point.py
+++ b/SLAM/iterative_closest_point/iterative_closest_point.py
@@ -44,7 +44,7 @@ def icp_matching(previous_points, current_points):
                 fig = plt.figure()
                 ax = fig.add_subplot(111, projection='3d')
                 ax.scatter(previous_points[0, :], previous_points[1, :],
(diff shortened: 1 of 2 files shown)


NobleBumblebee:
Hello, I have a failed check caused by another module, I wasn't working on.

AtsushiSakai:
Thank you for reporting. I will try to fix in #468. Please wait.

AtsushiSakai:
Rerun CI.

NobleBumblebee:
It's all good for the checks. Thank you for solving that issue.

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -152,6 +148,29 @@ def main():
         print("R:", R)
         print("T:", T)
 
+    # test for 3d point set
```
AtsushiSakai, line 151: I think this test for 3d point test should be separated to main(), like main_3d_point() or something. And then iterative_closest_point.py should be updated.

AtsushiSakai (changes requested):
Thank you for great PR!!. I have a small request. PTAL.

NobleBumblebee:
There is one thing that bothers me. Even if you pass 3d point cloud, the animation would be in 2d (for axis x,y). The code to visualize in 3d will be more complicated. Requiring to put if clause for 2d or 3d case, etc.
Is it ok, that we left 2d animation for both cases?

AtsushiSakai (changes requested):
Could you please add a new test here which call main_3d_points?
https://github.com/AtsushiSakai/PythonRobotics/blob/b3d39f5039aabbc45a4f85fa60096ae8a8bbbe01/tests/test_iterative_closest_point.py#L8

AtsushiSakai:
One idea is we can create a new plot function for 2d points to extract these codes:
https://github.com/AtsushiSakai/PythonRobotics/blob/b3d39f5039aabbc45a4f85fa60096ae8a8bbbe01/SLAM/iterative_closest_point/iterative_closest_point.py#L38-L47
And creating a new plot function for 3d points and dispatch to call which function using a if clause.

NobleBumblebee:
Can I ask you something? I am new to this style checking. How can I do these checks on my Spyder on Windows 10 before the commits? I always have some style issues.

NobleBumblebee:
So for 3d plots I divided 2d case and 3d case by if-else operator. To my mind it is least dirty.
It is possible to put drawing code to separate functions, but I am afraid it will not look good among functions that implement functionality.
What do you think?

NobleBumblebee:
Unused import concerns `from mpl_toolkits.mplot3d import Axes3D` for 3D plotting. It is needed for 3d scatter drawing but it is never called explicitly.

AtsushiSakai:
> How can I do these checks on my Spyder on Windows 10 before the commits? I always have some style issues.

Unfortunately, there is not a way not. But I think the "runtests.sh" should do it locally.
I will try to add it in #481

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -35,15 +36,24 @@ def icp_matching(previous_points, current_points):
...
             plt.gcf().canvas.mpl_connect(
                 'key_release_event',
                 lambda event: [exit(0) if event.key == 'escape' else None])
-            plt.plot(previous_points[0, :], previous_points[1, :], ".r")
-            plt.plot(current_points[0, :], current_points[1, :], ".b")
-            plt.plot(0.0, 0.0, "xr")
-            plt.axis("equal")
+            if previous_points.shape[0] == 3:
+                fig = plt.figure()
+                ax = fig.add_subplot(111, projection='3d')
```
AtsushiSakai, line 45: After initialization, the figure should be clean up like:
```suggestion
                plt.cla()
```

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -35,15 +36,24 @@ def icp_matching(previous_points, current_points):
         count += 1
```
AtsushiSakai, line 36: Before the while loop, you can initialize the figure like:
```
if show_animation:  # pragma: no cover
        fig = plt.figure()
        if previous_points.shape[0] == 3:
            ax = fig.add_subplot(111, projection='3d')

```

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -35,15 +36,24 @@ def icp_matching(previous_points, current_points):
...
 
         if show_animation:  # pragma: no cover
```
AtsushiSakai, line 38: Could you please extract these plot related code inside the if clause to a function like `plot_points`?

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -5,6 +5,7 @@
```
AtsushiSakai, line 5: If you want, you can add your name in the header comment.

AtsushiSakai (changes requested):
Hi. Thank you for your patience. Maybe this is the last review. After addressed my comments, I will merge this PR. PTAL.

NobleBumblebee:
I tried different options to make a 3d scatter be drawn on the same figure. None of them worked for me. PTAL.
Options I tried:
- just `plt.cla()` before drawing scatter
- no clearing at all
- `axes.clear()`
- All of the above with recreating of axes `axes = figure.add_subplot(111, projection='3d')`
- Using `figure.show()` at the end
- Using `figure.canvas.draw()  figure.canvas.flush_events()`
None of if was working for 3d plot. I left some code commented. That may help you. So I left working solution with recreating figure for 3d.

NobleBumblebee:
We also need to do something with import of `from mpl_toolkits.mplot3d import Axes3D` as it will continue to produce alerts.

SLAM/iterative_closest_point/iterative_closest_point.py
```diff
@@ -124,6 +118,31 @@ def svd_motion_estimation(previous_points, current_points):
...
 
+def plot_points(previous_points, current_points, figure):
+    # for stopping simulation with the esc key.
+    plt.gcf().canvas.mpl_connect(
+        'key_release_event',
+        lambda event: [exit(0) if event.key == 'escape' else None])
+    if previous_points.shape[0] == 3:
+        # axes = figure.axes[0]
+        # axes.clear()
+        figure = plt.figure()
```
AtsushiSakai, line 129: I'm not sure why are you creating figure every iteration?
This works without figure creation in my environment.
```suggestion
                plt.clf()
```

AtsushiSakai (changes requested):

AtsushiSakai:
One more thing. I think you can run style checker locally with:
1. pip install pycodestyle
2. run `././rundiffstylecheck.sh`
Please try it.

SLAM/iterative_closest_point/iterative_closest_point.py
NobleBumblebee: For 3d case in Spyder I can see only the first plot with initial point position. And it is not updating.
(Need to mention that for 2d case every single frame of optimization is displayed as a separate image, and not an updated previous)
May be this is the case for Spyder. Does it work for you without creating figure for 3d?

NobleBumblebee:
`bash rundiffstylecheck.sh` didn't worked for me:

> rundiffstylecheck.sh start!
> --2021-02-22` 15:27:06--  https://github.com/AtsushiSakai/DiffSentinel/archive/v0.1.6%0D.zip%0D
> 
> Resolving github.com (github.com)... 140.82.121.4
> Connecting to github.com (github.com)|140.82.121.4|:443... connected.
> HTTP request sent, awaiting response... 404 Not Found
> 2021-02-22 15:27:06 ERROR 404: Not Found.
> 
> .ZIP.or v0.1.6 find or open v0.1.6
> rundiffstylecheck.sh: line 6: ./DiffSentinel*/starter.sh: No such file or directory
> rundiffstylecheck.sh: line 16: syntax error: unexpected end of file

Calling `pycodestyle` could be used directly for every single file instead. #481
`pycodestyle --show-source --show-pep8 SLAM/iterative_closest_point/iterative_closest_point.py`

NobleBumblebee:
Hello @AtsushiSakai , still no answer from you.

> There is one thing that bothers me. Even if you pass 3d point cloud, the animation would be in 2d (for axis x,y). The code to visualize in 3d will be more complicated. Requiring to put if clause for 2d or 3d case, etc.
> Is it ok, that we left 2d animation for both cases?

As I said at the beginning 3d plots are complicated because of a lot of non evident moments.
I suggest to keep 2d plots for both cases.

SLAM/iterative_closest_point/iterative_closest_point.py
AtsushiSakai: Yeah. In my environment (just run the script from a terminal), it works. I think creating a lot of figures is annoying. Could you try to run the script from terminal?

AtsushiSakai:
I'm sorry I missed your comments.

> I suggest to keep 2d plots for both cases.
I think 3d plot is very useful for understanding what is happening in the script and other many examples already are using 3d plot.
If you don't mind, once this PR is merged, I will fix the 3d plot related code. Is it OK for you?

NobleBumblebee:
I commented the part of 3d plots, so that now it works good with 2d plots. I leave the 3d part for you to modify.

AtsushiSakai (approved):
LGTM. Thank you for your patience!!

NobleBumblebee:
Thank you too. Please keep me in touch with when you will add 3d plot drawing. I want to see a solution that works for you.

Issue #342, opened 2020-06-17
CVXPY cannot be installed


willzoe:
The development environment I currently use is Ubuntu16.04+Python3.8+anaconda+pycharm.

But when I install CVXPY in anaconda to the virtual environment, it prompts that it cannot be found.

What is the recommended development environment?

lentychang:
Hi @willzoe,  

try `conda install -c conda-forge cvxpy -n <ENV_NAME>` it should work, unless you meet some unsolved dependency.

or if conda doesn't work, you can just use pip to intall

AtsushiSakai:
@willzoe Hi please check install doc of cvxpy: https://www.cvxpy.org/install/
If it don't work, please report it for cvxpy team.

@lentychang  Hi. Thank you for your support. I applicate it.

PR #483, opened 2021-02-16
Fix: cvxpy cannot be installed

wang701:
I ran into #342 where I cannot just create a new conda environment using the existing `environment.yml` file.

#### Reference issue
See #342.

1. I added a channel to the yml file so that conda knows where to look for packages like `cvxpy`.
2. Since [requirements](https://pythonrobotics.readthedocs.io/en/latest/getting_started.html#requirements) needs Python to be 3.9 and up. I specified it in the yml file as well.

#### Additional information

None.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,9 +1,11 @@
 name: python_robotics
+channels:
+  - conda-forge
 dependencies:
-- python
-- pip
-- scipy
-- numpy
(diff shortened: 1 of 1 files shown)


AtsushiSakai (approved):
Good catch. Thank you!!

PR #521, opened 2021-06-29
Dubins path bug fix

SchwartzCode:
<!-- 
Thanks for contributing a pull request! 
Please ensure that your PR satisfies the checklist before submitting:
- [] This project only accept codes for python 3.9 or higher.
- [] If you add a new algorithm sample code, please add a unit test file under `test` dir.
     This sample test code might help you : https://github.com/AtsushiSakai/PythonRobotics/blob/master/tests/test_a_star.py
- [] If you fix a bug of existing code please add a test function in the test code to show the issue was solved.
- [] Please fix all issues on CI (All CI should be green), before code review.

#### Reference issue
None applicable

Bug in Dubin's path planner where the planner would occasionally take a step opposite the direction of the path if the change in x or y over the course of the entire path was zero (or super close to 0, like 4e-15). Adding the equals sign has fixed this in all of my test cases where this was happening. I can send some images over if you'd like :)

#### Additional information
I also made a test script for this file since none existed in /tests/

This project is awesome! Has helped me a ton this summer :1st_place_medal:


SchwartzCode:
Small bug where path would occasionally take a step in the wrong direction if change in x over the course of the path was 0 or close to it (i.e. 4e-15). This also happened with if change in y was very close to zero.

Pictures of my test cases (I'm using this code with another project) before and after making this change available upon request

PathPlanning/DubinsPath/dubins_path_planning.py
```diff
@@ -255,7 +255,7 @@ def generate_local_course(total_length, lengths, mode, max_curvature,
     ll = 0.0
 
     for (m, l, i) in zip(mode, lengths, range(len(mode))):
-        if l > 0.0:
+        if l >= 0.0:
```
AtsushiSakai, line 258: when l = 0.0, we don’t need to update path points.
so it does not work?
if l = 0.0:
    continue
elseif l > 0.0:
    ….

AtsushiSakai (commented):
Thank you for your PR.
I have a small question.
Also, could you fix CI error?

PathPlanning/DubinsPath/dubins_path_planning.py
SchwartzCode: Makes sense to me, I broke out l == 0 like you suggested. I also tried to appease the CI errors, will change more if that isn't happy still

SchwartzCode:
Okay looks like I managed to pass the CI checks, let me know if there is anything else I should do for this PR :)

AtsushiSakai (approved):
LGTM

Issue #523, opened 2021-07-01
Reeds shepp path has length 0 handling issue.


AtsushiSakai:
Reeds shepp path planner code has same issue of #521

PR #524, opened 2021-07-01
Issue #523 fix (Reeds-Shepp planner handling length=0 case)

SchwartzCode:
Addressing #523 by adding handling for length=0 case to Reeds-Shepp path planner

Sorry for it being so many commits, kept getting tripped up with merges


AtsushiSakai (approved):
LGTM. Thank you again!!.

PR #535, opened 2021-08-10
Fix markdown equations syntax

ksvbka:

Fix some small markdown equations syntax to display correctly. Thanks


AtsushiSakai (approved):
Make sense. Thanks.

PR #548, opened 2021-09-28
Modify notations

vltanh:
I modified some of the notations so that it will be less confusing (I was confused reading it).

- It is better to specify of which the Jacobian matrix is, here it is the transition and measurement function `f` and `g` (writing it as `J_F` and `J_H` is misleading)
- It is better to use `Delta` so as not to confuse people with the `d` commonly used in derivative
- It is more correct to use `\partial` instead of `d` since it is more multivariate than it is not
- Fix minor errors like extra brackets

### DIFF
--- a/docs/modules/extended_kalman_filter_localization.rst
+++ b/docs/modules/extended_kalman_filter_localization.rst
@@ -123,7 +123,7 @@ where
 
 The observation function states that
 
-:math: `\begin{equation*} \begin{bmatrix} x' \\ y' \end{bmatrix} = g(\textbf{x}) = \begin{bmatrix} x \\ y \end{bmatrix} \end{equation*}`
+:math:`\begin{equation*} \begin{bmatrix} x' \\ y' \end{bmatrix} = g(\textbf{x}) = \begin{bmatrix} x \\ y \end{bmatrix} \end{equation*}`
 
 Its Jacobian matrix is
 
--- a/docs/modules/extended_kalman_filter_localization.rst
+++ b/docs/modules/extended_kalman_filter_localization.rst
@@ -77,32 +77,36 @@ Motion Model
 
 The robot model is
 
-.. math::  \dot{x} = vcos(\phi)
+.. math::  \dot{x} = v \cos(\phi)
 
-.. math::  \dot{y} = vsin((\phi)
+.. math::  \dot{y} = v \sin(\phi)
 
 .. math::  \dot{\phi} = \omega
 
 So, the motion model is
 
-.. math:: \textbf{x}_{t+1} = F\textbf{x}_t+B\textbf{u}_t
+.. math:: \textbf{x}_{t+1} = f(\textbf{x}_t, \textbf{u}_t) = F\textbf{x}_t+B\textbf{u}_t
 
 where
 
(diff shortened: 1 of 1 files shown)


docs/modules/extended_kalman_filter_localization.rst
```diff
@@ -77,32 +77,36 @@ Motion Model
...
 
-:math:`dt` is a time interval.
+:math:`\Delta t` is a time interval.
 
 This is implemented at
 `code <https://github.com/AtsushiSakai/PythonRobotics/blob/916b4382de090de29f54538b356cef1c811aacce/Localization/extended_kalman_filter/extended_kalman_filter.py#L53-L67>`__
 
+The motion function is that
+
+:math: `\begin{equation*} \begin{bmatrix} x' \\ y' \\ w' \\ v' \end{bmatrix} = f(\textbf{x}, \textbf{u}) = \begin{bmatrix} x + v\cos(\phi)\Delta t \\ y + v\sin(\phi) \\ \phi + \omega \Delta t \\ v \end{bmatrix} \end{equation*}`
```
AtsushiSakai, line 103: Maybe this space is not needed.
```suggestion
:math:`\begin{equation*} \begin{bmatrix} x' \\ y' \\ w' \\ v' \end{bmatrix} = f(\textbf{x}, \textbf{u}) = \begin{bmatrix} x + v\cos(\phi)\Delta t \\ y + v\sin(\phi) \\ \phi + \omega \Delta t \\ v \end{bmatrix} \end{equation*}`
```

docs/modules/extended_kalman_filter_localization.rst
```diff
@@ -111,15 +115,19 @@ The robot can get x-y position infomation from GPS.
...
+.. math:: \textbf{z}_{t} = g(\textbf{x}_t) = H \textbf{x}_t
 
 where
 
-:math:`\begin{equation*} B= \begin{bmatrix} 1 & 0 & 0& 0\\ 0 & 1 & 0& 0\\ \end{bmatrix} \end{equation*}`
+:math:`\begin{equation*} H = \begin{bmatrix} 1 & 0 & 0 & 0 \\ 0 & 1 & 0 & 0 \\ \end{bmatrix} \end{equation*}`
+
+The observation function states that
+
+:math: `\begin{equation*} \begin{bmatrix} x' \\ y' \end{bmatrix} = g(\textbf{x}) = \begin{bmatrix} x \\ y \end{bmatrix} \end{equation*}`
```
AtsushiSakai, line 126: Maybe this space is not needed.
```suggestion
:math:`\begin{equation*} \begin{bmatrix} x' \\ y' \end{bmatrix} = g(\textbf{x}) = \begin{bmatrix} x \\ y \end{bmatrix} \end{equation*}`
```

AtsushiSakai: In my environment, this space might cause rendering issue:
SS 2021-10-03 at 16 49 13

AtsushiSakai (changes requested):
Thank you for your PR. I agree your notations look better. I have some commnets. PTAL.

vltanh:
I agree, it was because I am not familiar with this format so I thought it does not matter. Thank you for pointing that out.

AtsushiSakai (approved):
LGTM. Thanks.

PR #595, opened 2021-12-10
Add speed limitation of the robot

Muhammad-Yazdian:
I implemented the robot's speed limitations (i.e., maximum linear speed and maximum angular speed). 

There was not any open issue. I just did a minor improvement by adding 10 lines of codes.

### DIFF
--- a/Control/move_to_pose/move_to_pose.py
+++ b/Control/move_to_pose/move_to_pose.py
@@ -20,8 +20,8 @@ Kp_beta = -3
 dt = 0.01
 
 # Robot specifications
-_MAX_LINEAR_SPEED = 15
-_MAX_ANGULAR_SPEED = 7
+MAX_LINEAR_SPEED = 15
+MAX_ANGULAR_SPEED = 7
 
 show_animation = True
 
@@ -67,11 +67,11 @@ def move_to_pose(x_start, y_start, theta_start, x_goal, y_goal, theta_goal):
         if alpha > np.pi / 2 or alpha < -np.pi / 2:
             v = -v
 
-        if abs(v) > _MAX_LINEAR_SPEED:
-            v = v/abs(v) * _MAX_LINEAR_SPEED 
+        if abs(v) > MAX_LINEAR_SPEED:
+            v = v/abs(v) * MAX_LINEAR_SPEED 
 
-        if abs(w) > _MAX_ANGULAR_SPEED:
-            w = w/abs(w) * _MAX_ANGULAR_SPEED
+        if abs(w) > MAX_ANGULAR_SPEED:
+            w = w/abs(w) * MAX_ANGULAR_SPEED
 
         theta = theta + w * dt
         x = x + v * np.cos(theta) * dt
--- a/Control/move_to_pose/move_to_pose.py
+++ b/Control/move_to_pose/move_to_pose.py
@@ -19,6 +19,10 @@ Kp_alpha = 15
 Kp_beta = -3
 dt = 0.01
 
+# Robot specifications
+_MAX_LINEAR_SPEED = 15
+_MAX_ANGULAR_SPEED = 7
+
 show_animation = True
 
 
@@ -62,6 +66,12 @@ def move_to_pose(x_start, y_start, theta_start, x_goal, y_goal, theta_goal):
 
         if alpha > np.pi / 2 or alpha < -np.pi / 2:
             v = -v
+
+        if abs(v) > _MAX_LINEAR_SPEED:
+            v = v/abs(v) * _MAX_LINEAR_SPEED 
+
+        if abs(w) > _MAX_ANGULAR_SPEED:
+            w = w/abs(w) * _MAX_ANGULAR_SPEED
 
         theta = theta + w * dt
         x = x + v * np.cos(theta) * dt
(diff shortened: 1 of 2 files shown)


Control/move_to_pose/move_to_pose.py
```diff
@@ -19,6 +19,10 @@
 Kp_beta = -3
 dt = 0.01
 
+# Robot specifications
+_MAX_LINEAR_SPEED = 15
+_MAX_ANGULAR_SPEED = 7
```
AtsushiSakai, line 24: Could you please remove the first _ for the two global configurations?

Control/move_to_pose/move_to_pose.py
```diff
@@ -63,6 +67,12 @@ def move_to_pose(x_start, y_start, theta_start, x_goal, y_goal, theta_goal):
         if alpha > np.pi / 2 or alpha < -np.pi / 2:
             v = -v
 
+        if abs(v) > _MAX_LINEAR_SPEED:
+            v = v/abs(v) * _MAX_LINEAR_SPEED
```
AtsushiSakai, line 71: Why you don’t just set maximum speed?

Control/move_to_pose/move_to_pose.py
```diff
@@ -63,6 +67,12 @@ def move_to_pose(x_start, y_start, theta_start, x_goal, y_goal, theta_goal):
         if alpha > np.pi / 2 or alpha < -np.pi / 2:
             v = -v
 
+        if abs(v) > _MAX_LINEAR_SPEED:
+            v = v/abs(v) * _MAX_LINEAR_SPEED 
+
+        if abs(w) > _MAX_ANGULAR_SPEED:
+            w = w/abs(w) * _MAX_ANGULAR_SPEED
```
AtsushiSakai, line 74: Could you please add a test for your addition? https://github.com/AtsushiSakai/PythonRobotics/blob/master/tests/test_move_to_pose.py

AtsushiSakai (changes requested):

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: We don't set the speed to max speed since we want to keep the control law alive. Note that when the robot gets to target (or very close to it) the robot speed should approaches zero.

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: OK. it is done.

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: Should I add another file with the name like 'test_move_to_pose_with_speed_limit.py'? Doesn't the current test do the same thing?

Control/move_to_pose/move_to_pose.py
AtsushiSakai: Sorry I don’t understand it well… Could elaborate it?
v is velocity, so it should be under the specification, right?
> Note that when the robot gets to target (or very close to it) the robot speed should approaches zero.

Yes. In that case, the speed must be under specification, your code does not work, right?

Control/move_to_pose/move_to_pose.py
AtsushiSakai: Adding a new test in the file is OK. You can change the two specifications by changing the global variables, then the control results should be changed. Checking it in the test is OK.

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: Sure. The actual name of 'v' should have been 'command_linear_velocity' which is the velocity the controller commands the robot to move. When far from the target, rho and thus v are high. Since in reality any robot has a max speed, we need to limit the 'command_linear_velocity' to the robot's max speed. How to categorize the variables (robot specification vs controller parameters) really depends on how we look at the system. I would separate controller and plant (i.e., the robot) as different subsystem modules and therefore It is better to keep v as a controller parameter than the robot specification.

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: I included the test_2 in the original test file.

Control/move_to_pose/move_to_pose.py
AtsushiSakai: OK. Thank you. I understood. Could you please use ‘np.sign’ instead of ‘v/-abs(v)’?

tests/test_move_to_pose.py
```diff
@@ -6,6 +6,15 @@ def test_1():
     m.show_animation = False
     m.main()
 
+def test_2():
+    """
+    This unit test tests the move_to_pose.py program for a random
```
AtsushiSakai, line 11: ‘random’ is no needed?

AtsushiSakai (changes requested):
Could you please fix CI failures?

50
./Control/move_to_pose/move_to_pose.py:71:44: W291 trailing whitespace
51
./tests/test_move_to_pose.py:9:1: E302 expected 2 blank lines, found 1
52
./tests/test_move_to_pose.py:11:66: W291 trailing whitespace
53
Error: Your changes contain pycodestyle errors.

Control/move_to_pose/move_to_pose.py
Muhammad-Yazdian: You are welcome. Yes, that is a good idea. I just updated to np.sign().

tests/test_move_to_pose.py
Muhammad-Yazdian: I removed it.

AtsushiSakai (approved):

AtsushiSakai:
Thanks, this is good enhancement. I will merge it.

PR #645, opened 2022-03-01
Fix a broken gif link in the "Lidar to grid map"

i-aki-y:
The gif link is broken now. I think this is the right URL.

#### Reference issue

no

doc only.

#### CheckList
-[] Did you add an unittest for your new example or defect fix?
-[] Did you add documents for your new example?
-[] All CIs are green? (You can check it after submitting)


AtsushiSakai (approved):
LGTM. Thank you!!

PR #647, opened 2022-03-10
Fix typo: correction for missing 'delta t' in EKF documentation

HrushikeshBudhale:
Reference issue
typo

What does this implement/fix?
Update motion function in localization/EKF documentation. Add missing delta t.


AtsushiSakai (approved):
Good catch. LGTM

Issue #700, opened 2022-08-01
bug in grid_map_lib


jsbyysheng:
**Describe the bug**
It is possible that the position `(iox, ioy)` to be checked with `iox == min_x`. This will cause an incorrect inside check.
<img width="198" alt="CleanShot 2022-08-02 at 00 11 56@2x" src="https://user-images.githubusercontent.com/8376541/182193473-5cbdc44a-b80e-4cc1-a0cc-f12178b7b72c.png">
This is the outline (the blue one) for grid map creation.
<img width="198" alt="CleanShot 2022-08-02 at 00 12 10@2x" src="https://user-images.githubusercontent.com/8376541/182193824-c7cdd7dc-5b4a-4745-8a73-9419ab7cef19.png">
Because of `iox == min_x`, an incorrect pixel bar occur.
<img width="197" alt="CleanShot 2022-08-02 at 00 12 24@2x" src="https://user-images.githubusercontent.com/8376541/182194276-95dbc343-4de6-48a2-a7ab-ff7075ff6eb4.png">
This mistake will cause the failure of the grid map based path planning.

jsbyysheng:
After #701 , we can get the result below.
<img width="170" alt="CleanShot 2022-08-02 at 00 18 39@2x" src="https://user-images.githubusercontent.com/837
<img width="168" alt="CleanShot 2022-08-02 at 00 18 50@2x" src="https://user-images.githubusercontent.com/8376541/182196390-b01720bd-7519-4b71-9054-a57e4896c2fa.png">
6541/182196361-267c8c42-2dda-4bd4-a146-fc710a28aec6.png">

PR #701, opened 2022-08-01
fix bugs in check a position inside the polygon

jsbyysheng:
It is possible that the position `(iox, ioy)` to be checked with `iox == min_x`. This will cause a incorrect inside check. #700


AtsushiSakai (commented):
Thank you for your PR. Is it possible to add a test for the boundary condition [here](https://github.com/AtsushiSakai/PythonRobotics/blob/master/tests/test_grid_map_lib.py)?

jsbyysheng:
A boundary condition is added to the test.
CleanShot 2022-08-07 at 10 50 18@2x
Without this patch, we can see the error result.
CleanShot 2022-08-07 at 10 51 23@2x
With this patch.

AtsushiSakai (approved):

PR #748, opened 2022-11-13
Added PT-BR translation for README.

OtavioRMC:


AtsushiSakai:
Thank you for your PR. However, I'm sorry I cannot maintain a non-English document in the future... So if you believe the translation is needed for many users, can you maintain your forked repository? I'm sorry.

Issue #755, opened 2022-11-17
The robot doesn't reach the target velocity in frenet_optimal_trajectory.py


KohMat:
**Describe the bug**
The target speed is defined as constant in `frenet_optimal_trajectory.py`:
https://github.com/AtsushiSakai/PythonRobotics/blob/8f14488ec2f243e48222bc87923604a7d46c45af/PathPlanning/FrenetOptimalTrajectory/frenet_optimal_trajectory.py#L40

However, as shown in the GIF, the robot doesn't reach the target speed during the simulation period.

![](https://github.com/AtsushiSakai/PythonRoboticsGifs/raw/master/PathPlanning/FrenetOptimalTrajectory/animation.gif)

**Expected behavior**
The robot should reach the target speed.

**Additional context**
I think that the acceleration in the start state is not set correctly  in the longitudinal trajectory generation.
The current code sets the acceleration to 0:
https://github.com/AtsushiSakai/PythonRobotics/blob/8f14488ec2f243e48222bc87923604a7d46c45af/PathPlanning/FrenetOptimalTrajectory/frenet_optimal_trajectory.py#L142

However, looking at the paper, the correct start state is the current robot acceleration:
![start_state](https://user-images.githubusercontent.com/13505518/202418159-c945f7c2-7e50-49d1-90ab-cc94cab05ccc.png)

I changed the code to use the current acceleration and got the result that the robot almost reached the target velocity.

```python
lon_qp = QuarticPolynomial(s0, c_speed, c_accel, tv, 0.0, Ti)
```

![frenet_trajectory](https://user-images.githubusercontent.com/13505518/202421184-0feb2186-ce43-4c17-afc2-f215156498a6.gif)

AtsushiSakai:
Thank you for your report. Can you create a PR?

PR #756, opened 2022-11-17
Fix the start state to use the current accel in frent planner(#755)

KohMat:
#### Reference issue
#755

The robot didn't reach the target velocity when running frenet_optimal_trajectory.py.

To fix this, I changed the constraints for determining polynomials that generate longitudinal trajectories according to paper.

Expressly, I set the initial acceleration to the current acceleration.

#### CheckList


AtsushiSakai (approved):
LGTM. Thank you!!

PR #764, opened 2022-12-25
Undefined name: import sys for line 188

cclauss:
#### CheckList

### DIFF
--- a/PathPlanning/RRT/rrt_with_sobol_sampler.py
+++ b/PathPlanning/RRT/rrt_with_sobol_sampler.py
@@ -28,6 +28,7 @@ Rojas (rafaelrojasmiliani@gmail.com)
 
 import math
 import random
+import sys
 import matplotlib.pyplot as plt
 import numpy as np
 


PathPlanning/RRT/rrt_with_sobol_sampler.py
```diff
@@ -28,6 +28,7 @@
 
 import math
 import random
+import sys
```
AtsushiSakai, line 31: Thank you for your PR. Is it possible to add a test to run this script?

cclauss: https://github.com/AtsushiSakai/PythonRobotics/blob/master/tests/test_rrt_with_sobol_sampler.py

AtsushiSakai: OK…. There is, I’m no sure why the test did not detect the missing import.

AtsushiSakai: OK. This code will be used only when path drawing. Make sense

AtsushiSakai (approved):
LGTM. Thanks

PR #766, opened 2022-12-29
add axes3d to figure

zlingkang:
The Bipedal Planner code won't show the drawing. The reason is that since Matplotlib V3.4.0, Axes3D automatically adding itself to Figure is deprecated. The fix here is to add the axes3d to figure manually.

#### Additional information
Reference reading: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.4.0.html


AtsushiSakai (approved):
LGTM. Thanks

PR #767, opened 2022-12-30
Utilize numpy to reduce calculation time

kadirhas:
Change-Id: I6e421a1c2524a3d8f8875121a1a6d2ed832c3150

I've noticed that the current Implementation uses lists. By replacing them with numpy arrays and using built in numpy functions such as any, I reduced the calculation time of d_star_lite. The other thing that I've noticed, once after initializing class, if the user wants to add new spoofed objects and calculate a new path, it was not using the prior information that is generated. So I modified the initialization process a bit to use the prior information.
#### CheckList

### DIFF
--- a/PathPlanning/DStarLite/d_star_lite.py
+++ b/PathPlanning/DStarLite/d_star_lite.py
@@ -53,8 +53,6 @@ class DStarLite:
         Node(-1, -1, math.sqrt(2))
     ]
 
-    initialized = False
-
     def __init__(self, ox: list, oy: list):
         # Ensure that within the algorithm implementation all node coordinates
         # are indices in the grid and extend
@@ -73,19 +71,14 @@ class DStarLite:
         self.U = list()  # type: ignore
         self.km = 0.0
         self.kold = 0.0
-        self.rhs = np.full(
-            (self.x_max, self.y_max),
-            float("inf")  # type: ignore
-        )
-        self.g = np.full(
-            (self.x_max, self.y_max),
-            float("inf")  # type: ignore
-        )
+        self.rhs = self.create_grid(float("inf"))
+        self.g = self.create_grid(float("inf"))
         self.detected_obstacles = list()  # type: ignore
         self.detected_obstacles_xy = np.array([])
         if show_animation:
             self.detected_obstacles_for_plotting_x = list()  # type: ignore
             self.detected_obstacles_for_plotting_y = list()  # type: ignore
+        self.initialized = False
 
     def create_grid(self, val: float):
         return np.full((self.x_max, self.y_max), val)
@@ -98,11 +91,11 @@ class DStarLite:
             & (self.obstacles_xy[:, 1] == y)
         ).any()
         is_in_detected_obstacles = False
+
         if self.detected_obstacles_xy.shape[0] > 0:
-            is_in_detected_obstacles = (
-                (self.detected_obstacles_xy[:, 0] == x)
-                & (self.detected_obstacles_xy[:, 1] == y)
-            ).any()
+            is_x_equal = self.detected_obstacles_xy[:, 0] == x
+            is_y_equal = self.detected_obstacles_xy[:, 1] == y
+            is_in_detected_obstacles = (is_x_equal & is_y_equal).any()
 
         return is_in_obstacles or is_in_detected_obstacles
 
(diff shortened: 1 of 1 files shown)


kadirhas:
The styling warnings are caused by W503, which says that it is [going to change](https://www.flake8rules.com/rules/W503.html).

PathPlanning/DStarLite/d_star_lite.py
```diff
@@ -167,23 +187,25 @@ def compare_keys(self, key_pair1: tuple[float, float],
...
-                                 self.calculate_key(self.start))) or \
-                self.rhs[self.start.x][self.start.y] != \
-                self.g[self.start.x][self.start.y]:
+        while (
+            len(self.U) > 0
+            and self.compare_keys(self.U[0][1], self.calculate_key(self.start))
+        ) or (
+            self.rhs[self.start.x, self.start.y]
+            != self.g[self.start.x, self.start.y]
+        ).any():
```
AtsushiSakai, line 196: Not this PR topic, but I think each while condition is challenging to read, so is it possible to introduce boolean variables or other kinds of intermediate values for readability?

PathPlanning/DStarLite/d_star_lite.py
```diff
@@ -62,32 +65,46 @@ def __init__(self, ox: list, oy: list):
...
+        y = np.array([node.y])
+        is_in_obstacles = (
+            (self.obstacles_xy[:, 0] == x)
+            & (self.obstacles_xy[:, 1] == y)
+        ).any()
+        is_in_detected_obstacles = False
+        if self.detected_obstacles_xy.shape[0] > 0:
+            is_in_detected_obstacles = (
+                (self.detected_obstacles_xy[:, 0] == x)
+                & (self.detected_obstacles_xy[:, 1] == y)
```
AtsushiSakai, line 104: Difficult to read each condition, is it possible to improve it?

PathPlanning/DStarLite/d_star_lite.py
```diff
@@ -62,32 +65,46 @@ def __init__(self, ox: list, oy: list):
...
-        self.g = list()  # type: ignore
+        self.rhs = np.full(
+            (self.x_max, self.y_max),
+            float("inf")  # type: ignore
+        )
+        self.g = np.full(
+            (self.x_max, self.y_max),
+            float("inf")  # type: ignore
+        )
         self.detected_obstacles = list()  # type: ignore
```
AtsushiSakai, line 84: self.dtected_obstacles is still needed?

PathPlanning/DStarLite/d_star_lite.py
```diff
@@ -62,32 +65,46 @@ def __init__(self, ox: list, oy: list):
...
         self.goal = Node(0, 0)
         self.U = list()  # type: ignore
         self.km = 0.0
         self.kold = 0.0
-        self.rhs = list()  # type: ignore
-        self.g = list()  # type: ignore
+        self.rhs = np.full(
+            (self.x_max, self.y_max),
+            float("inf")  # type: ignore
+        )
```
AtsushiSakai, line 79: I think introducing a new function to initialize a grid map is better like:

```
def init2dGrid():
       return np.full(self.x_max, self.y_max), float("inf")
```

PathPlanning/DStarLite/d_star_lite.py
```diff
@@ -52,6 +53,8 @@ class DStarLite:
         Node(-1, -1, math.sqrt(2))
     ]
 
+    initialized = False
```
AtsushiSakai, line 56: This should be initizlied in the `__init__`?

AtsushiSakai (changes requested):
Thank you for your PR. I have some points to improve the readability of this code. PTAL.

PathPlanning/DStarLite/d_star_lite.py
kadirhas: That is a great suggestion, thank you! I've updated this as you have suggested. Please take a look.

PathPlanning/DStarLite/d_star_lite.py
kadirhas: I separated the conditions with boolean variables.

PathPlanning/DStarLite/d_star_lite.py
kadirhas: It is still used in random object generation. I will change that function as well and test it.

PathPlanning/DStarLite/d_star_lite.py
kadirhas: I used an already existing function called create_grid. Thanks for the improvement.

PathPlanning/DStarLite/d_star_lite.py
kadirhas: Yes, I moved it, thank you

PathPlanning/DStarLite/d_star_lite.py
kadirhas: I removed it, and while doing that fixed a bug about going out of range of the rhs.

AtsushiSakai (approved):
LGTM. Thank you for great PR!!

PR #834, opened 2023-05-15
fix: set rrt_star to find the shortest path

### LINKED ISSUE #833: RRT Star is not choosing the path of the lowest cost.
**Describe the bug**

RRT Star is not choosing the path of the lowest cost.  Here is the related code in this function. 
https://github.com/AtsushiSakai/PythonRobotics/blob/b877a9d5ca15b794a72221c33d52d13bf3bf8434/PathPlanning/RRTStar/rrt_star.py#L166-L169
It tries to find a node around the goal such that  `start->...->node->goal` is the shortest path. However, it actually returns the node around the goal with **_the lowest cost from starting point to the node_**, instead of what we want: **_the lowest cost from starting point to goal_**.

Here is an example from one of my projects. It shows that the cost of the best paths is not always decreasing.

image

**Expected behavior**
It should choose the node that forms a shortest path from start point to goal.

**Screenshots**
None

**Desktop (please complete the following information):**
 - Python version: 3.9.9
 - Each library version: install from requirements.txt
 - OS version: MacOS 13.0.1

TKONIY:
#### Reference issue
fix #833 

I fix the bug that RRT Star is not choosing the shortest path. Details of the bug are illustrated in #833. 
I did not post any tests because I thought it is obvious. Hope you understand.
#### CheckList

### DIFF
--- a/PathPlanning/RRTStar/rrt_star.py
+++ b/PathPlanning/RRTStar/rrt_star.py
@@ -163,9 +163,13 @@ class RRTStar(RRT):
         if not safe_goal_inds:
             return None
 
-        min_cost = min([self.node_list[i].cost for i in safe_goal_inds])
-        for i in safe_goal_inds:
-            if self.node_list[i].cost == min_cost:
+        safe_goal_costs = [self.node_list[i].cost + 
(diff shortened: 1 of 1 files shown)


AtsushiSakai (approved):
LGTM. Thank!!

PR #890, opened 2023-08-15
fix dstar

### LINKED ISSUE #886: Bug report: D*
**Describe the bug**
D* can't cope with a changed map: When I add new obstacles after the start node closes, the program gets stuck.

    def run(self, start, end):
       
        rx = []
        ry = []

        self.insert(end, 0.0)

        while True:
            self.process_state()
            if start.t == "close":
                break

        start.set_state("s")
        s = start
        s = s.parent
        s.set_state("e")
        tmp = start

        # here I add new obstacles
        ox, oy = [], []
        for i in range(15, 21):
            ox.append(i)
            oy.append(40)
        self.map.set_obstacle([(i, j) for i, j in zip(ox, oy)])
        plt.pause(0.001)
        plt.plot(ox, oy, ".k")

        while tmp != end:
            tmp.set_state("*")
            rx.append(tmp.x)
            ry.append(tmp.y)
            if show_animation:
                plt.plot(rx, ry, "-r")
                plt.pause(0.01)
            if tmp.parent.state == "#":
                self.modify(tmp)
                continue
            tmp = tmp.parent
        tmp.set_state("e")

        return rx, ry

**Expected behavior**
D* Dynamically adjust the path.

**Solution**
change the 'elif' in process_state function to 'if'

    def process_state(self):

          *****
  
          if k_old < x.h:
              for y in self.map.get_neighbors(x):
                  if y.h <= k_old and x.h > y.h + x.cost(y):
                      x.parent = y
                      x.h = y.h + x.cost(y)
          if k_old == x.h: # change elif to if
              for y in self.map.get_neighbors(x):
                  if y.t == "new" or y.parent == x and y.h != x.h + x.cost(y) \
                          or y.parent != x and y.h > x.h + x.cost(y):
                      y.parent = x
                      self.insert(y, x.h + x.cost(y))
                      if plot_y:
                          plt.pause(time_interval)
                          plt.plot(y.x, y.y, 'xg')

         ******

**PS**
I don't know whether this is the correct solution of this bug, in https://www.cs.cmu.edu/~motionplanning/lecture/AppH-astar-dstar_howie.pdf, page 36, it is 'if' rather than 'elif'.

Ts-sound:
#### Reference issue
fix #886 

When you modify the map ( add obstacle ) , the untraveled paths are updated ；

### DIFF
--- a/PathPlanning/DStar/dstar.py
+++ b/PathPlanning/DStar/dstar.py
@@ -11,6 +11,8 @@ import math
 
 from sys import maxsize
 
+import time
+
 import matplotlib.pyplot as plt
 
 show_animation = True
@@ -55,7 +57,7 @@ class Map:
         self.map = self.init_map()
 
     def init_map(self):
-        map_list = []
+        map_list:list[State] = []
         for i in range(self.row):
             tmp = []
             for j in range(self.col):
@@ -86,7 +88,7 @@ class Map:
 
 class Dstar:
     def __init__(self, maps):
-        self.map = maps
+        self.map:Map = maps
         self.open_list = set()
 
     def process_state(self):
@@ -172,6 +174,8 @@ class Dstar:
         s = s.parent
         s.set_state("e")
         tmp = start
+        
+        return tmp
 
         while tmp != end:
             tmp.set_state("*")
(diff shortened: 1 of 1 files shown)


2892510130:
Thanks for your PR, but there is still a little bug here: when your newly added obstacle is too long it will not work. I think the reason is that the initialization of Map is [0 -> 100], so it does not initialize points have negative x coordinate. However the obstacle, ox, have negative value. Maybe we should change the Map initialization code?

2892510130:
Another question: I read your code, and I find that you rerun the D* in a cloned map. Is it still a D* algorithm? D* should be abled to dynamically find a path based on previous search result, but your dstar2 is a new search without information from a previous search.

2892510130:
I've fixed the bug, but there's still a problem with the map initialization process (it's not a problem with the algorithm, so I don't want to waste time fixing it, haha). #896

PR #896, opened 2023-08-22
fix D* bug, #886

2892510130:
D* now can dynamically change the path when new obstacles are added.


PathPlanning/LPAStar/lpa_star.py
```diff
@@ -0,0 +1,284 @@
+"""
+Lifelong Planning A*
+author: flyingmars (github.com/2892510130)
```
AtsushiSakai, line 3: Please make a PR focusing on one issue. Do not include multiple contributions in one PR.

Issue #944, opened 2023-12-01
Strange Animation Results in test_rrt_star_reeds_shepp.py


videh25:
### Bug
test3 in test_rrt_star_reeds_shepp.py tries to generate a path with step_size = 20. Since it is a test file, animation is kept off. However, on turning the animation on, the result is very  weird. RRT does not plan with smooth RS curves but rather with straight lines with sharp corners.

### Expected Behaviour
Also ran test2 from test_rrt_star_reeds_shepp.py (step_size = 0.2). They plan and generate smooth RS curves only. 
 
### Screenshots
**test3 Animation**
![Screenshot from 2023-12-01 17-02-51](https://github.com/AtsushiSakai/PythonRobotics/assets/66770479/6242cfc5-dc76-4100-b5bc-e9b158792ae9)

***test2 Animation***
![Screenshot from 2023-12-01 17-10-18](https://github.com/AtsushiSakai/PythonRobotics/assets/66770479/2f2f0dcd-7ee3-47bf-8d15-be3450dbc506)

### Desktop
 - Python 3.9.5
 - numpy 1.24.3
 - scipy 1.10.1
 - matplotlib 3.7.1
 - cvxpy 1.4.1
 - Ubuntu 20.04

AtsushiSakai:
@videh25 Is it OK to close this issue by your PR?

videh25:
Yes, sounds great!

PR #892, opened 2023-08-17
Added: missing path types

videh25:
fix #341 

Added the missing RS paths in reeds_shepp_path_planning.py as mentioned in the issue. 
Changed the function {generate_path} a make the code more neat.
Removed these functions {curve_curve_curve, curve_straight_curve, straight_curve_straight}: To have a neat implementation and remove inaccurate path type.
Added these functions{reflect, timeflip}: To neatly consider the reflected and timeflipped paths as mentions in the paper.

Referred Paper: OPTIMAL PATHS FOR A CAR THAT GOES BOTH FORWARDS AND BACKWARDS - J. A. REEDS AND L. A. SHEPP [There were some formula typos in the paper]
Referred Code: https://github.com/nathanlct/reeds-shepp-curves/blob/master/reeds_shepp.py#L139

#### CheckList
- [ ] Did you add an unittest for your new example or defect fix? - NA
- [ ] Did you add documents for your new example? - NA
- [x] All CIs are green? (You can check it after submitting)

PR #949, opened 2023-12-04
Reopen: Added missing path types

### LINKED ISSUE #341: Add missing reeds_shepp curve type in  reeds_shepp_path_planning.py
why there is curve type SLS  from function " straight_left_straight(x, y, phi) ". I read the paper but this type doesn't exist. 
Does this code  implement the algorithm completely? maybe it lacks the last five base words in the picture 
image

videh25:
fix #341 

Reopening the PR #892 since it was closed due to deletion of head repository.

Committed the same changes:
1. Added missing RS paths and refactored the code.
2. Modified assert condition in test3 of test_rrt_star_reeds_shepp.py.

Summarising the discussion from PR #892:
1. "x" in the function names refers to the direction switches.
2. Had to change the assert condition in test3 of test_rrt_star_reeds_shepp.py due to some strange behaviour. I have mentioned this in detail in  this [issue](https://github.com/AtsushiSakai/PythonRobotics/issues/944).

#### CheckList


PathPlanning/ReedsSheppPath/reeds_shepp_path_planning.py
```diff
@@ -110,114 +83,193 @@ def left_straight_left(x, y, phi):
...
-            return True, t, u, v
+            return True, [t, u, v], ['L', 'S', 'L']
 
-    return False, 0.0, 0.0, 0.0
+    return False, [], []
 
 
-def left_right_left(x, y, phi):
-    u1, t1 = polar(x - math.sin(phi), y - 1.0 + math.cos(phi))
+def left_straight_right(x, y, phi):
```
AtsushiSakai, line 91: Is it possible to update the doc to describe mathematics for each path type?
https://github.com/AtsushiSakai/PythonRobotics/blob/master/docs/modules/path_planning/reeds_shepp_path/reeds_shepp_path_main.rst

tests/test_rrt_star_reeds_shepp.py
```diff
@@ -39,9 +39,7 @@ def test3():
                                              max_iter=100, step_size=step_size)
     rrt_star_reeds_shepp.set_random_seed(seed=8)
     path = rrt_star_reeds_shepp.planning(animation=False)
-    for i in range(len(path)-1):
-        # + 0.00000000000001 for acceptable errors arising from the planning process
-        assert m.math.dist(path[i][0:2], path[i+1][0:2]) < step_size + 0.00000000000001
```
AtsushiSakai, line 44: Why this code should be changed?

AtsushiSakai (changes requested):

PathPlanning/ModelPredictiveTrajectoryGenerator/trajectory_generator.py
```diff
@@ -18,7 +18,7 @@
 
 # optimization parameter
 max_iter = 100
-h = np.array([0.5, 0.02, 0.02]).T  # parameter sampling distance
+h: np.ndarray = np.array([0.5, 0.02, 0.02]).T  # parameter sampling distance
```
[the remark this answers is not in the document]
videh25: test_mypy_type_check.py fails without the annotation.

PathPlanning/ReedsSheppPath/reeds_shepp_path_planning.py
videh25: Sure... will do it soon

tests/test_rrt_star_reeds_shepp.py
videh25: This specific test seems to be a bit off. I have explained this in detail in [issue #994](https://github.com/AtsushiSakai/PythonRobotics/issues/944).

PathPlanning/ReedsSheppPath/reeds_shepp_path_planning.py
videh25: I have updated the documentation... Please take a look into it.

AtsushiSakai: Oh, great document!!!. Thank you!!

image

tests/test_rrt_star_reeds_shepp.py
AtsushiSakai: It looks like this test just check the path point interval length just less than `step_size`.
You mean this test starts failing after this PR changes?

PathPlanning/ModelPredictiveTrajectoryGenerator/trajectory_generator.py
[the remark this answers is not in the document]
AtsushiSakai: Hmm. I’m not sure why this annotation is needed… Because type is obvious… But it is OK, not harmful.

tests/test_rrt_star_reeds_shepp.py
videh25: Even if the test passes with the original code, the paths generated are not appropriate as explained in the issue. However, with this PR's changes, the test also fails. 

Here is a possible explanation I have for this:

1. `calc_interpolate_dists_list` function in reeds_shepp_path_planning.py uses `interp_dists = np.arange(0.0, length, d_dist)` in line 328 to generate steps for a given path length. So if `length < step_length` then `interp_dists = [0]`. 

2. Older code has a path_type: straight_curve_straight, which is absent in the PR version.  Generally, straight maneuvers are greater then step length but curved maneuvers are not. So curved paths end up getting skipped in the path returned by reeds_shepp_path_planning. And what we see are just straight maneuvers appended to each other.

3. The newer code has does not have the path type and is supposed to return shorter paths. Thus, any path returned by newer code is node able to have a maneuver greater than step_size (20 units). However, the new code passes the test with `step_size = 0.2`.

Thus, either the step_size should be reduced appropriately or no path should be expected with the current step_size.

AtsushiSakai: OK. Make sense. I agree when the `step_size` is too big, the function should return None instead of invalid path points.
Did you add any validation for the too big `step_size`?
And, could you please change test function name from `test3` to `test_too_big_step_size` or something?

videh25: I can add the validation if required. Will do the renaming.

AtsushiSakai (approved):
LGTM. Thank you!!

PR #994, opened 2024-03-21
Update ndt_map.py

xiao-ge4:
#### Reference issue
The original file uses self._create_grid_index_map(ox, oy) before self.grid_map, causing a logical conflict

The initialization of self.grid_map is brought forward before the create NDT grid index map

### DIFF
--- a/Mapping/ndt_map/ndt_map.py
+++ b/Mapping/ndt_map/ndt_map.py
@@ -52,6 +52,8 @@ class NDTMap:
         center_y = np.mean(oy)
         self.ox = ox
         self.oy = oy
+        # init grid_map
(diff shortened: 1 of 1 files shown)


Mapping/ndt_map/ndt_map.py
```diff
@@ -52,14 +52,15 @@ def __init__(self, ox, oy, resolution):
         center_y = np.mean(oy)
         self.ox = ox
         self.oy = oy
+        # init grid_map
+        self.grid_map = GridMap(width, height, resolution, center_x, center_y, self.NDTGrid())
```
AtsushiSakai, line 56: This change is harmless I think. But, could you please let me know, why this change is needed? What kind of issue happened in your environment?

PR #1000, opened 2024-04-08
fix::Control PathFinderController bug

noob3-3:
The PathFinderController class in Control/move_to_pose/move_to_pose.py, line 77 Kp_beta, should be self. Not the controller.


AtsushiSakai (approved):

AtsushiSakai:
@noob3-3 Thank you!!

PR #1018, opened 2024-05-14
fixed hard-coded plot limits

yousefbilal:
The limits of the plots were fixed to (-2,15) which is the example random range. Adjusted to be rand_min, rand_max instead.


AtsushiSakai:
@yousefbilal Thank you for your PR. Is it possible to post animation movies before and after changes?

yousefbilal:
Sure, this is the plot before the change. Some parts are hidden because the plot limits are [-2,15]
image

After the change, which sets the limits of the plot to the random sample area
image

Note that this also affects RRTStar as it extends the RRT class.

AtsushiSakai:
@yousefbilal Thank you!!

PR #1042, opened 2024-07-02
Python sample codes compatible with AirSim Unreal PyMavLink and MavProxy

ss48:

I am adding extra files that make it compatible with AirSim | Unreal | PyMavLink | MavProxy | ardupilot
#### CheckList
- [ y ] Did you add documents for your new example?


AtsushiSakai:
Sorry too big PR, I cannot review this. Please create an issue to propose this change, then split this PR.

PR #1075, opened 2024-09-12
Performance improvement

AryazE:
#### Reference issue
None

This improves performance by transforming `calculated` into a set. As the only use is to check if a value is already calculated, keeping each value once is enough and it is faster to search in a set than a list.

#### Additional information
This issue was found by [DyLin](https://github.com/AryazE/DyLin) via [auto-dylin](https://github.com/AryazE/auto-dylin).

#### CheckList


AtsushiSakai (approved):
LGTM

PR #1086, opened 2024-10-20
Fix: Hybrid A* direction is incorrect

parmaski:
#### Reference issue
None

Hybrid A* outputs path containing x, y, yaw, and direction.
But direction is not preserved correctly and its size is smaller than x, y, yaw.

##### before fix
size of x, y, yaw, direction -> 1235, 1235, 1235, 510

##### after fix

size of x, y, yaw, direction -> 1235, 1235, 1235, 1235

 - Red line: Path.direction_list=True (forward)
 - Blue line: Path.direction_list=False (backward)

#### Additional information

`hybrid_a_star.py` throws module not found error because its dependency `reeds_shepp_path_planning.py` uses a module in root directory (`utils/angle.py`).
Therefore `sys.path` is also modified following other algorithms depending on `utils/angle.py`.

#### CheckList


AtsushiSakai (approved):
LGTM. Thanks!!

PR #1168, opened 2025-02-19
Update move_to_pose for cases where alpha > pi/2 or alpha < -pi/2

### LINKED ISSUE #877: move_to_pose.py not consistent with book when alpha > pi/2
**Describe the bug**
As the book says,
> For the case where the goal is behind the robot, that is α ∉ (−pi/2 , pi/2 ], we reverse the vehicle by negating v and γ in the control law

But the implementation here just negate linear velocity.

**Expected behavior**
I tested with negating both linear velocity & angular velocity, the control vibrates, but the overall trajectory makes more sense than only negating the linear velocity.

**Screenshots**
1. Negate Both
negate_both

2. Negate only linear velocity
negate_one

Aglargil:
#### Reference issue

fix #877

- Update move_to_pose for cases where alpha > pi/2 or alpha < -pi/2
- Format move_to_pose.py using Ruff

#### Additional information

Current results:
img_v3_02jl_e948e03d-773c-431a-81ec-7e74972d29ag

Previous results(master):
img_v3_02jl_5a226f00-9e2f-4b64-9cad-507cb12cc26g

#### CheckList
- [x] Did you add an unittest for your new example or defect fix?

### DIFF
--- a/Control/move_to_pose/move_to_pose.py
+++ b/Control/move_to_pose/move_to_pose.py
@@ -5,6 +5,7 @@ Move to specified pose
 Author: Daniel Ingram (daniel-s-ingram)
         Atsushi Sakai (@Atsushi_twi)
         Seied Muhammad Yazdian (@Muhammad-Yazdian)
+        Wang Zheng (@Aglargil)
 
 P. I. Corke, "Robotics, Vision & Control", Springer 2017, ISBN 978-3-319-54413-7
 
@@ -12,8 +13,13 @@ P. I. Corke, "Robotics, Vision & Control", Springer 2017, ISBN 978-3-319-54413-7
 
 import matplotlib.pyplot as plt
 import numpy as np
-from random import random
+import sys
+import pathlib
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent.parent))
 from utils.angle import angle_mod
+from random import random
+
 
 class PathFinderController:
     """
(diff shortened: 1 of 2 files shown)


Control/move_to_pose/move_to_pose.py
```diff
@@ -71,13 +77,18 @@ def calc_control_command(self, x_diff, y_diff, theta, theta_goal):
...
 
         rho = np.hypot(x_diff, y_diff)
-        alpha = angle_mod(np.arctan2(y_diff, x_diff) - theta)
-        beta = angle_mod(theta_goal - theta - alpha)
         v = self.Kp_rho * rho
-        w = self.Kp_alpha * alpha - self.Kp_beta * beta
 
+        alpha = angle_mod(np.arctan2(y_diff, x_diff) - theta)
+        beta = angle_mod(theta_goal - theta - alpha)
         if alpha > np.pi / 2 or alpha < -np.pi / 2:
```
AtsushiSakai, line 84: Do you think you can add a test case for this?

AtsushiSakai (commented):
Thank you!!. Adding a test case is great, but if it is difficult, it is ok to merge. let me know.

Control/move_to_pose/move_to_pose.py
Aglargil: To add test cases, I made move_to_pose() return the trajectory information executed, which is used to determine the stability of the execution and whether the endpoint is reached

AtsushiSakai (approved):
LGTM. Thanks!! @Aglargil

PR #1221, opened 2025-05-24
Add root directory to system path for lattice planner

YutharsanS:
#### Reference issue
Couldn't run lattice_planner.py because it was giving module not found error.

It adds the root of the directory to the sys.path so that eliminates the module not found error.

#### CheckList


AtsushiSakai (approved):
LGTM

PR #1300, opened 2025-10-21
new aco algorithm

anishk85:
This PR introduces the Ant Colony Optimization (ACO) algorithm to our path-planning module.

Instead of a single "searcher," this algorithm simulates a whole colony of virtual ants. They wander from the start to the goal, leaving a "pheromone" trail behind them. Shorter paths get reinforced faster (since ants complete the trip more often), and old trails slowly "evaporate."

Over time, the colony's collective intelligence naturally converges on the shortest and most efficient path. It's a super cool, nature-inspired approach!
animation


anishk85:
@AtsushiSakai please do have a look and please let me know if you need something from my side

